Issue #3, opened 2026-02-03
examples and schema validation use profile_name vs profileName


vasu2856:
The error still says to use profileName (not profile_name). The CLI code is reading profileName from the manifest but the schema validation is rejecting it. This is a bug in the 
CLI.

vasu2856:
The fix is merged. Resolving the issue.

PR #4, opened 2026-02-03
Fixing the profile name casing

vasu2856:
*Issue #, if available:*
https://github.com/aws/CICD-for-SageMakerUnifiedStudio/issues/3

*Description of changes:*
Fixing the profile name casing to switch to camel case as expected by code. As part of the change, updating 1. manifest  schema 2. manifests 


abaror75 (commented):
I used to have validation regression test that suppose to run validation on all manifests , why did it not catch it?   Can we make sure there is no additional issues like this ?

Shnekit (dismissed):
LGTM

vasu2856:
> I used to have validation regression test that suppose to run validation on all manifests , why did it not catch it? Can we make sure there is no additional issues like this ?

have removed legacy manifest schema and refactored some of the tests. Do take a look

abaror75 (approved):

PR #49, opened 2026-02-19
fix: describe command no longer requires AWS credentials without --connect

Shnekit:
- Add resolve_aws_pseudo_vars flag to substitute_env_vars() and load_yaml() to skip STS calls for AWS_ACCOUNT_ID/STS_ACCOUNT_ID/STS_REGION when not needed
- Thread flag through validate_manifest_file(), validate_yaml_syntax(), and ApplicationManifest.from_file()
- describe command only resolves AWS pseudo vars and domain names via live AWS calls when --connect is explicitly specified
- Without --connect, falls back to os.getenv for pseudo vars, leaves placeholders as-is if not set, and shows 'Unknown' for tag-based domains
- Fix domain name display to be per-stage instead of using first stage for all
- Add comma-separated list support for --targets option
- Fix bootstrap actions display to respect --targets filter

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/smus_cicd/application/application_manifest.py
+++ b/src/smus_cicd/application/application_manifest.py
@@ -239,7 +239,9 @@ class ApplicationManifest:
     _file_path: Optional[str] = field(default=None, init=False)
 
     @classmethod
-    def from_file(cls, manifest_file: str, resolve_aws_pseudo_vars: bool = True) -> "ApplicationManifest":
+    def from_file(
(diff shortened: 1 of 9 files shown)


vasu2856 (commented):
Will need to update the unit and integration tests

Shnekit:
> Will need to update the unit and integration tests

Updated, please check again @vasu2856 :)

vasu2856 (approved):
LGTM

PR #60, opened 2026-02-25
Add DataZone catalog import/export spec (requirements, design, tasks)

vasu2856:
*Description of changes:*

Specs to add support for Catalog resources - assets, assettypes, formtypes, glossarryterms and glossaries. The spec also highlights special handling of schedule asset. 


abaror75:
SageMakerUnifiedStudioScheduleAssetType - That perhaps is the least interesting when you talk about Catalog..  hide this as implementation detail

abaror75:
Lets create user manual first.   I think requirement 1 is not that critical .. Its more important to allow the user to define a set of data products, or a set of assets, and have the tool find all the connected entities (BG, FormTypes) and export them first , so the import will work with all dependent objects no matter what.  Check Quick import to see how they did it.   

Start with the simplest user experience to use it ..  Also , I would like to see example on how we get that timestamp to put into the CLI when I use a GitHub Pipeline ,  how will we figure this out?  Should we add tags to Git when doing the import ?  Lets figure out entire user experience here.

.kiro/specs/datazone-catalog-import-export/design.md
```diff
@@ -0,0 +1,479 @@
...
+    glossaries:
+      include:              # optional, defaults to both
+        - glossaries
+        - glossaryTerms
+      updatedAfter: "2025-01-01T00:00:00Z"  # optional ISO 8601 filter
+    dataProducts:
+      names:                # optional, defaults to all data products
+        - "Sales Analytics Product"
+        - "Customer Insights Product"
+      updatedAfter: "2025-01-01T00:00:00Z"  # optional ISO 8601 filter
```
Shnekit: Hmmm, wondering if we can read this dynamically from the previously created bundle date. I think this is what customers are going to do manually anyway

abaror75:
1. Focus on Assets and Data Products.   Explain dependent entities as carried along.   
2. Discuss ownership of assets for the project and that only owned assets can be exported and imported 
Manifest focus too much on EntityTypes.. I think this can be one filter .. why replicate ..  Why I don't have assetFilter by name ,  by physical type (glue table, model) ?

abaror75:
3. Manifest should not contain updateAfter date.  That should be option .. otherwise manifest need to keep changing

abaror75:
3. Manifest should not contain updateAfter date.  That should be option .. otherwise manifest need to keep changing

abaror75:
Assets (reference AssetTypes)
Glossaries (no dependencies)
GlossaryTerms (reference Glossaries)
Data Products (may reference Assets)

How can you do Assets before Glossaries ?

.kiro/specs/datazone-catalog-import-export/design.md
```diff
@@ -0,0 +1,479 @@
...
+
+```yaml
+content:
+  catalog:
+    assets:
+      include:              # optional, defaults to all three
+        - formTypes
+        - assetTypes
+        - assets
+      updatedAfter: "2025-01-01T00:00:00Z"  # optional ISO 8601 filter
```
DaviesHuang: What's the use case of `updatedAfter`? When is it updated?

.kiro/specs/datazone-catalog-import-export/design.md
vasu2856: yup. Was exploring ways to do it. There are scripts you are run as part of Github workflow, but since we don't manage the state, i don't think we can do it on the CLI directly.

abaror75:
Can I do dry-run to make sure all the things that need to be in the target are there before I make changes.. I don't want partial import failure

abaror75:
I think we need two filters  :  Name filter and EntityType filter , default *:*

.kiro/specs/datazone-catalog-import-export/design.md
vasu2856: Its used to limit the scope of export to what's changed since last deployment. Omitting it will pull everything in the source project.

abaror75:
We already discussed having the deploy add Git tag to resources.. maybe along the lines for that , we add Git Tag during the bundle ( generate bundle Id and store it) , then use that for the deploy .  Or the user provide these timestamps on the command line , but again , wonder how this works.  The doc does not describe it

.kiro/specs/datazone-catalog-import-export/design.md
Shnekit: We can. I think I am going to add bundle versions deployed to targets S3 as part of rollback support :)

abaror75:
Publish Listing feature

.kiro/specs/datazone-catalog-import-export/design.md
```diff
@@ -0,0 +1,535 @@
...
+
+### Property 11: Dependency-Ordered Creation
+**Validates: Requirement 5.6**
+
+For all import operations, the `CatalogImporter` SHALL invoke create APIs such that: every Glossary is created before any GlossaryTerm that references it, every GlossaryTerm is created before any Asset or FormType that references it, every FormType is created before any AssetType that references it, and every AssetType is created before any Asset that references it.
+
+### Property 12: Dependency-Ordered Deletion
+**Validates: Requirement 5.4, 5.5**
+
+For all import operations where resources exist in the target project but are NOT present in the bundle, the `CatalogImporter` SHALL invoke delete APIs in reverse dependency order: Assets before AssetTypes, AssetTypes before FormTypes, GlossaryTerms before Glossaries (to avoid breaking references).
```
Shnekit, line 382: So, this currently exists for DZ catalogs only, not any other resource, right?

Shnekit: I think we will need to communicate to the customer that they should not deploy DZ catalogs using this tool to any environment which engineers have write access to. This will be a drawback for some use-cases

.kiro/specs/datazone-catalog-import-export/design.md
```diff
@@ -0,0 +1,535 @@
...
+
+### Property 13: Import Error Resilience
+**Validates: Requirements 5.10, 5.14, 7.3**
+
+For any import operation where `K` out of `N` resources fail during create/update/delete/publish API calls (where `0 < K < N`), the `CatalogImporter` SHALL still attempt to process all `N` resources, log each of the `K` failures with resource name, type, and error message, and report a summary containing the failure count.
```
Shnekit, line 387: How hard is it to treat all resources as a single transaction? So that we can delete the already created resources in case of failure, instead of relying on the customer to clean-up all the mess after us.

.kiro/specs/datazone-catalog-import-export/design.md
```diff
@@ -0,0 +1,535 @@
...
+
+### Property 14: Import Summary Counts
+**Validates: Requirement 5.12, 6.3**
+
+For all import operations, the `CatalogImporter` SHALL return counts `{created, updated, deleted, failed, published}` where `created + updated + deleted + failed` equals the total number of resources processed (resources in bundle + resources in target not in bundle), and `published` equals the number of assets and data products that were published in the source (listingStatus == "ACTIVE") and whose listing was verified as ACTIVE in the target when skipPublish is false.
```
Shnekit, line 392: Assuming the count will make more sense to the customer once we implement dry-run to actually show how many resources we planned to deploy vs how many we actually deployed

.kiro/specs/datazone-catalog-import-export/design.md
vasu2856: its not a transaction, as they are individual API calls. We can't assume that the rollback strategy is always deletion, as some of the resources could already exist.  We are introducing the dry run feature next which will run by default for every deployment to detect any potential failures and stop deployment. Customer would be able to redeploy on failure.

.kiro/specs/datazone-catalog-import-export/design.md
vasu2856: not sure if I follow the relation to engineer's write access. This is a feature we are building and its documented here - https://github.com/vasu2856/CICD-for-SageMakerUnifiedStudio-public/blob/feature/catalog_import_export/docs/catalog-import-export-guide.md#dependency-order

.kiro/specs/datazone-catalog-import-export/design.md
vasu2856: Technically, it shows how many failed right now. Dry run will show a similar output with details on the failures, but without making the change.

.kiro/specs/datazone-catalog-import-export/requirements.md
```diff
@@ -0,0 +1,140 @@
...
+15. WHEN importing assets that contain a `DataSourceReferenceForm` in their `formsInput`, THE Catalog_Importer SHALL remap the form's `dataSourceIdentifier.id`, `filterableDataSourceId`, and `dataSourceRunId` to a matching data source in the target project. The target data source SHALL be matched by type and database name: (1) exact `databaseName` match in the data source's `relationalFilterConfigurations`, (2) wildcard `"*"` database filter, (3) first candidate of the same type. The `databaseName` SHALL be extracted from the asset's `GlueTableForm` content. If no matching data source exists in the target project, the `DataSourceReferenceForm` SHALL be stripped from the asset's forms.
+
+### Requirement 6: Deploy Command Integration
+
+**User Story:** As a developer, I want catalog import to be integrated into the existing deploy command flow, so that catalog resources are deployed alongside other bundle content.
+
+#### Acceptance Criteria
+
+1. WHEN the deploy command processes a bundle containing a `catalog/catalog_export.json` file, THE Deploy_Command SHALL invoke the Catalog_Importer after storage and QuickSight deployments
```
Shnekit, line 125: Can there be any dependencies between the storage or QuickSight and DZ catalogs? If so, does the order always respects these dependencies?

.kiro/specs/datazone-catalog-import-export/requirements.md
```diff
@@ -0,0 +1,140 @@
...
+2. WHERE the stage's `deployment_configuration.catalog.disable` is set to true, THE Deploy_Command SHALL skip catalog import for that stage
+3. THE Deploy_Command SHALL report the count of created, updated, and failed catalog resources upon completion
+4. IF all catalog resource imports fail, THEN THE Deploy_Command SHALL report the deployment as failed
```
Shnekit, line 128: Why? Should we mark it as failed when any of the resources failed to deploy?

docs/architecture.md
```diff
@@ -39,37 +39,24 @@ The SMUS CI/CD CLI is a command-line tool that automates the deployment of data
...
-┌─────────────────────────────────────────────────────────────┐
-│                    Integration Layer                         │
-│  (AWS Service Helpers: DataZone, S3, Glue, SageMaker, etc.) │
-└─────────────────────────────────────────────────────────────┘
-                            ↓
-┌─────────────────────────────────────────────────────────────┐
-│                      AWS Services                            │
-│  (DataZone, S3, MWAA, Glue, SageMaker, Athena, etc.)       │
-└─────────────────────────────────────────────────────────────┘
+```mermaid
```
Shnekit, line 48: Fancy :) 

Might need different color choices to make it more readable though
Image

docs/architecture.md
```diff
@@ -85,67 +72,100 @@ Bootstrap actions use a registry pattern for extensibility:
 
 ### High-Level System Diagram
 
+```mermaid
+graph TB
+    subgraph DevWorkstation[Developer Workstation]
```
Shnekit: These come out white and completely unreadable

docs/architecture.md
```diff
@@ -85,67 +72,100 @@ Bootstrap actions use a registry pattern for extensibility:
...
+- **Flexibility**: Different organizational units can own different domains
+- **Cross-Account Support**: Stages can span multiple AWS accounts
+- **Independent Governance**: Each domain can have its own governance policies and access controls
+
+**Configuration Example:**
+
+```yaml
+stages:
+  dev:
+    domain_id: dzd_dev123456  # Development domain
```
Shnekit: This struct is incorrect. It should be 
```
prod:
    domain:
      id: dzd_dev123456
    project:
      name: my-app-dev
```

docs/architecture.md
```diff
@@ -283,11 +306,13 @@ src/smus_cicd/helpers/
 **Key Helpers:**
 
 1. **DataZone Helper** (`datazone.py`):
-   - Resolve domain IDs from tags
-   - Create/update projects
+   - Resolve domain IDs from tags or explicit domain_id configuration
```
Shnekit: ... domain `id` ...

docs/catalog-import-export-guide.md
```diff
@@ -0,0 +1,542 @@
...
+| **AssetTypes** | Asset type definitions that specify structure and metadata for categories of assets |
+| **Assets** | Catalog assets representing data resources with metadata and forms |
+| **Data Products** | Data products that bundle one or more data assets for publishing and sharing |
+
+## Key Assumption: Physical Resources Must Have the Same Name
+
+> ⚠️ **IMPORTANT**: Catalog import/export assumes that the underlying physical resources (e.g., Glue Tables, Glue Databases, S3 buckets) referenced by your catalog assets **have the same name in both source and target environments**. The import process matches assets across environments using normalized `externalIdentifier` values or resource names. If the physical resource names differ between environments (e.g., `my-table-dev` vs `my-table-prod`), the matching will fail and resources will be created as new rather than updated. Ensure your infrastructure provisioning uses consistent resource names across stages, or that your naming conventions allow the normalization logic (which strips AWS account IDs and region strings) to produce matching identifiers.
```
Shnekit: We should also mention here that if the names do not match, it will not only be created from scratch, but also the resource that does not match will be deleted, right?

docs/catalog-import-export-guide.md
```diff
@@ -0,0 +1,542 @@
...
+
+1. **Configure your manifest** with `content.catalog.enabled: true`
+2. **Run bundle** to export catalog resources from your source project
+3. **Run deploy** to import catalog resources into your target project
+
+### Example: Full Deployment Pipeline
+
+```bash
+# Step 1: Bundle from dev environment (exports ALL catalog resources)
+smus-cli bundle --manifest manifest.yaml
```
Shnekit: You will need to rebase carefully.

I see you are referencing `smus-cli` in some places and `smus-cicd-cli` in others and neither of them are correct...

docs/catalog-import-export-guide.md
```diff
@@ -0,0 +1,542 @@
...
+smus-cli deploy --bundle bundle.zip --targets prod
+```
+
+### Example: Incremental Export with --updated-after
+
+To export only recently modified resources, use the `--updated-after` CLI flag on the bundle command. This filters ALL resource types uniformly by modification timestamp.
+
+```bash
+# Export only resources modified since February 1, 2025
+smus-cli bundle --manifest manifest.yaml --updated-after "2025-02-01T00:00:00Z"
```
Shnekit: Hmmm, are we applying this `--updated-after` feature to all resources now (not just DZ catalogs)? If not, that will be very confusing...

docs/catalog-import-export-guide.md
```diff
@@ -0,0 +1,542 @@
...
+smus-cli deploy --bundle bundle.zip --targets test
+```
+
+## Resource Mapping and Dependencies
+
+### Identifier Mapping
+
+The import process uses `externalIdentifier` (with normalization) as the primary mapping key between source and target projects. When a resource doesn't have an `externalIdentifier`, the `name` field is used as a fallback.
+
+- **externalIdentifier**: Normalized by removing AWS account ID and region information before matching
```
Shnekit, line 168: In theory, we can also support cross-partitioning by removing and replacing partitions as well, but I guess that will not be a very common use-case

docs/catalog-import-export-guide.md
```diff
@@ -0,0 +1,542 @@
...
+- **name**: Used as fallback when `externalIdentifier` is not present
+
+When a matching resource exists in the target project, it is updated. When no match is found, a new resource is created.
+
+> ⚠️ **IMPORTANT**: This mapping assumes that the underlying physical resources (e.g., Glue Tables, S3 buckets) have the **same name** across environments. For example, if a Glue Table is named `analytics_db.customers` in your dev environment, it must also be named `analytics_db.customers` in your test and prod environments for the asset to be correctly matched and updated rather than duplicated.
```
Shnekit, line 173: We are setting everything right, if we are deploying to an empty environment, right? We can also mention that here

docs/catalog-import-export-guide.md
```diff
@@ -0,0 +1,542 @@
...
+During import, resources that exist in the target project but are NOT present in the bundle are deleted. Deletion follows reverse dependency order to avoid breaking references:
+
+1. Data Products
+2. Assets
+3. AssetTypes
+4. FormTypes
+5. GlossaryTerms
+6. Glossaries
+
+## Export JSON Structure
```
Shnekit, line 209: Is this the JSON after the bundle step? Why are we providing the structure to the customer, if they are not supposed to view/edit the assets generated by bundle command?

docs/catalog-import-export-guide.md
```diff
@@ -0,0 +1,542 @@
...
+1. The error is logged with resource name, type, and error message
+2. Processing continues with remaining resources
+3. A summary reports counts of created, updated, deleted, and failed resources
+
+**Example output:**
+```
+Catalog import summary:
+  Created: 15
+  Updated: 8
+  Deleted: 2
```
Shnekit, line 304: Does the customer have any way of knowing what exactly was created/updated/deleted? I know the stuff is not arnable, but we should at least provide the names of resources (especially if we are deleting stuff)

docs/catalog-import-export-guide.md
```diff
@@ -0,0 +1,542 @@
...
+
+Commit your `manifest.yaml` with catalog export configuration to version control. This ensures catalog definitions are versioned alongside your application code.
+
+### 2. Test in Lower Environments First
+
+Always test catalog changes in dev/test environments before promoting to production:
+
+```bash
+# Deploy to dev first
+smus-cli deploy --bundle bundle.zip --targets dev
```
Shnekit: This goes back to the question of whether customers are supposed to deploy to dev, if it destroys all custom changes not included in the manifest.
We should probably specify that it must be an isolated empty project here

docs/catalog-import-export-guide.md
```diff
@@ -0,0 +1,542 @@
...
+
+# Verify catalog resources in dev
```
Shnekit: Nit: we can add a `smus-cli test` command here for that

docs/catalog-import-export-guide.md
```diff
@@ -0,0 +1,542 @@
...
+- Transient API throttling or service errors for specific resources
+- The target project lacks permissions for certain resource types
+
+**How to debug**:
+1. Re-run the deploy with verbose logging enabled:
+   ```bash
+   SMUS_LOG_LEVEL=DEBUG smus-cicd deploy --target <target-name>
+   ```
+2. Look for `Failed to import <resourceType> <name>:` lines in the output — these show the exact API error for each failed resource
+3. Check if failures cascade from a parent type:
```
Shnekit, line 478: How are we showing this to the customer? Can they tell by looking GlossaryTerms failure that the problem is actually in importing Glossaries?

docs/catalog-import-export-guide.md
```diff
@@ -0,0 +1,542 @@
...
+### Problem: Published count is 0 even though source assets were published
+
+**Cause**: The `create_listing_change_set` API is asynchronous. The import process polls the resource after publishing to verify the listing becomes ACTIVE. If the underlying physical resource (e.g., Glue Table) doesn't exist in the target environment, the listing will fail asynchronously and be counted as a publish failure rather than a success.
+
+**Solution**:
+- Verify the physical resources (Glue Tables, S3 buckets, etc.) exist in the target environment with the same names as in the source
+- Check the deploy logs for "Listing FAILED" or "Listing verification timed out" messages
+- If the physical resources can't be created in the target, use `skipPublish: true` in the manifest to skip publishing
+
+**Solution**:
```
Shnekit, line 493: Another solution? :D

docs/catalog-import-export-quick-reference.md
```diff
@@ -0,0 +1,221 @@
+# Catalog Import/Export Quick Reference
```
Shnekit: Is this needed? Skimmed through it, it provides the same content as catalog-import-export-guide. I think I would rather have customers go through a more verbose guide than have them ask questions after reading the reference which are answered in the guide.

examples/catalog-import-export/app_tests/test_catalog_deployed.py
```diff
@@ -0,0 +1,128 @@
...
+        return resp.get("items", [])
+    except Exception as e:
+        pytest.skip(f"Could not search {search_scope}: {e}")
+        return []
+
+
+def test_glossaries_exist(datazone_client, domain_id, project_id):
+    """Verify that glossaries were imported into the target project."""
+    items = _search_resources(datazone_client, domain_id, project_id, "GLOSSARY")
+    assert len(items) > 0, "No glossaries found in target project"
```
Shnekit, line 51: Why not fetch the exact number from the manifest?

examples/catalog-import-export/app_tests/test_catalog_deployed.py
```diff
@@ -0,0 +1,128 @@
...
+
+def test_assets_exist(datazone_client, domain_id, project_id):
+    """Verify that assets were imported into the target project."""
+    items = _search_resources(datazone_client, domain_id, project_id, "ASSET")
+    assert len(items) > 0, "No assets found in target project"
+    print(f"✅ Found {len(items)} asset(s) in target project")
+
+
+def test_asset_names_match(datazone_client, domain_id, project_id):
+    """Verify that expected asset names are present in the target project."""
```
Shnekit: This is checking whether asset names list is not empty. Not the actual names matching

examples/catalog-import-export/app_tests/test_catalog_deployed.py
```diff
@@ -0,0 +1,128 @@
...
+
+    published_count = 0
+    for item in items:
+        asset_item = item.get("assetItem", item.get("assetListing", {}))
+        # Check for listing or published indicators
+        listing_id = asset_item.get("listingId")
+        if listing_id:
+            published_count += 1
+
+    print(f"✅ {published_count}/{len(items)} assets have listing IDs (published)")
```
Shnekit: What is this testing? This one seems like it never fails

examples/catalog-import-export/app_tests/test_catalog_deployed.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def test_form_types_exist(datazone_client, domain_id, project_id):
+    """Verify that custom form types were imported into the target project."""
```
Shnekit: Same with this one

examples/catalog-import-export/manifest.yaml
```diff
@@ -0,0 +1,41 @@
+applicationName: CatalogImportExportExample
+content:
+  catalog:
+    enabled: true
+stages:
+  dev:
+    stage: DEV
+    domain:
```
Shnekit, line 8: This will break if we decide to add another domain for any reason, so better to add a 
```
tags:
        purpose: smus-cicd-testing
```

examples/catalog-import-export/manifest.yaml
```diff
@@ -0,0 +1,41 @@
...
+      region: ${DEV_DOMAIN_REGION:us-east-1}
+    project:
+      name: dev-catalog-project
+      owners:
```
Shnekit, line 14: Can we add `arn:aws:iam::${AWS_ACCOUNT_ID}:role/Admin` as another owner, so that we can check the deployment results manually?

examples/catalog-import-export/manifest.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - arn:aws:iam::${AWS_ACCOUNT_ID}:role/GitHubActionsRole-SMUS-CLI-Tests
+    deployment_configuration:
+      catalog: {}
+  test:
+    stage: TEST
+    domain:
+      region: ${TEST_DOMAIN_REGION:us-east-1}
+    project:
+      name: test-catalog-project
+      create: true
```
Shnekit: Will this fail?

examples/catalog-import-export/manifest.yaml
```diff
@@ -0,0 +1,41 @@
+applicationName: CatalogImportExportExample
+content:
+  catalog:
+    enabled: true
+stages:
+  dev:
+    stage: DEV
+    domain:
+      region: ${DEV_DOMAIN_REGION:us-east-1}
```
Shnekit: I think I removed the region defaults from all examples. Let's remove it from here as well

examples/catalog-import-export/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+No filter options (`include`, `names`, `assetTypes`, `updatedAfter`, etc.) exist in the manifest.
+
+## Filtering with `--updated-after`
+
+The **only** way to filter exported catalog resources is via the `--updated-after` CLI flag on the `bundle` command. This flag accepts an ISO 8601 timestamp and filters **all** resource types uniformly by their `updatedAt` timestamp.
+
+```bash
+# Export only resources modified after a specific date
+smus-cli bundle --manifest examples/catalog-import-export/manifest.yaml \
```
Shnekit: This will need `--targets` flag, right?

.github/workflows/catalog-import-export.yml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  deploy:
+    uses: ./.github/workflows/smus-bundle-deploy.yml
+    secrets:
+      AWS_ROLE_ARN_DEV: ${{ secrets.AWS_ROLE_ARN_DEV }}
+      AWS_ROLE_ARN_TEST: ${{ secrets.AWS_ROLE_ARN_TEST }}
+      AWS_ROLE_ARN_PROD: ${{ secrets.AWS_ROLE_ARN_PROD }}
+    with:
+      manifest_path: 'examples/catalog-import-export/manifest.yaml'
```
Shnekit, line 30: This will fail if DEV does not have DZ catalog already. We will need to run seed_catalog_data.py before running the workflow

examples/catalog-import-export/seed_catalog_data.py
```diff
@@ -0,0 +1,281 @@
...
+    try:
+        resp = client.create_glossary(
+            domainIdentifier=domain_id,
+            owningProjectIdentifier=project_id,
+            name=name,
+            description=description,
+            status="ENABLED",
+        )
+        print(f"  ✅ Created glossary: {name} (id: {resp['id']})")
+        return resp["id"]
```
Shnekit: Curious: what is this id? I thought these resources use names as ids or will that return the name?

examples/catalog-import-export/seed_catalog_data.py
```diff
@@ -0,0 +1,281 @@
...
+            summary["created"] += 1
+        else:
+            summary["skipped"] += 1
+
+    # --- Summary ---
+    print(f"\n{'='*50}")
+    print(f"🌱 Seed Summary:")
+    print(f"   Created: {summary['created']}")
+    print(f"   Skipped: {summary['skipped']} (already exist)")
+    print(f"   Failed:  {summary['failed']}")
```
Shnekit: This is not getting set anywhere, this will always be 0

src/smus_cicd/application/application_manifest.py
```diff
@@ -283,38 +319,83 @@ def from_dict(cls, data: Dict[str, Any]) -> "ApplicationManifest":
...
+                        asset = AssetConfig(
+                            selector=selector,
+                            permission=asset_data.get("permission", "READ"),
+                            requestReason=asset_data.get(
+                                "requestReason", "Required for pipeline deployment"
+                            ),
+                        )
+                        access_assets.append(asset)
+
+                # Old location: accessAssets (backward compatibility)
```
Shnekit: Curious: do we need that for any of our examples or is it for customers? Since we have not really launched the thing yet, can we remove it?

src/smus_cicd/helpers/catalog_import.py
```diff
@@ -0,0 +1,1467 @@
...
+
+
+def _identify_resources_to_delete(
+    client,
+    domain_id: str,
+    project_id: str,
+    catalog_data: Dict[str, Any],
+) -> Dict[str, List[Dict[str, str]]]:
+    """
+    Find resources in target project that are not present in the bundle.
```
Shnekit, line 1078: Wait, I am confused. What happens when you specify `--updated-after` flag in the bundle command? I thought in this case bundle will contain the subset of resources listed in the manifest (only the ones that were updated). But we still want the untouched resources listed in the manifest to be there. Are we deleting them as per this logic?

src/smus_cicd/cli.py
```diff
@@ -158,6 +158,11 @@ def bundle(
         "--local",
         help="Bundle from local filesystem instead of dev target",
     ),
+    updated_after: str = typer.Option(
+        None,
+        "--updated-after",
```
Shnekit: Maybe we should name this `--update-catalog-after` to not make the customers think that the feature is supported for everything under the manifest

src/smus_cicd/cli.py
```diff
@@ -158,6 +158,11 @@ def bundle(
...
+        help="ISO 8601 timestamp to filter ALL catalog resources uniformly by updatedAt (e.g., 2024-01-01T00:00:00Z). This is a CLI-only option, not a manifest field.",
```
Shnekit: Nit: we can remove `This is a CLI-only option, not a manifest field.`

src/smus_cicd/cli.py
```diff
@@ -169,6 +174,26 @@ def bundle(
     """Create bundle zip files. By default bundles FROM dev target for deployment to specified target."""
     configure_logging(output, LOG_LEVEL)
 
+    # Validate --updated-after timestamp format if provided
+    if updated_after is not None:
+        from datetime import datetime
+
+        try:
```
Shnekit: We can also add a check here, if the provided time is in the future to fail fast

tests/integration/catalog-import-export/test_catalog_edge_cases.py
```diff
@@ -0,0 +1,362 @@
...
+                catalog_files = [
+                    n for n in zf.namelist() if n.startswith("catalog/")
+                ]
+                assert len(catalog_files) == 0, (
+                    f"Bundle should NOT contain catalog/ directory when "
+                    f"catalog.enabled is false. Found: {catalog_files}"
+                )
+        else:
+            # Bundle failed (no files to bundle) — also valid when catalog
+            # is the only content type and it's disabled.
```
Shnekit: We should validate that catalog is actually the only resource in the manifest. Otherwise, we are just allowing the test to fail with any other reason as well

tests/integration/catalog-import-export/test_catalog_edge_cases.py
```diff
@@ -0,0 +1,362 @@
...
+        self.cleanup_test_directory()
+
+    # ------------------------------------------------------------------
+    # Helpers
+    # ------------------------------------------------------------------
+
+    def _manifest_path(self, name: str) -> str:
+        return os.path.join(os.path.dirname(__file__), name)
+
+    def _get_project_info(self, pipeline_file: str, stage: str = "dev"):
```
Shnekit: Can we define these methods and some common variables in the helper file somewhere to avoid code repetition?

Shnekit (commented):
We can sync to review the comments. There are some very broad ones I posted which I need clarification on

.kiro/specs/datazone-catalog-import-export/requirements.md
vasu2856: So technically, some of the glue jobs or workflows could be populate the catalog. Hence the the catalog resources import is setup to be after storage and quick sight dashboards. If the jobs had a dependency on catalog resources, customer will have to wait for deploy to be successful and then run the job.

.kiro/specs/datazone-catalog-import-export/requirements.md
vasu2856: fair point. Let me update the requirements  to fail the deployment when there is atleast one failure. With the dry-run PR, the implementation is changing to run dry-run first and then fail the deployment. Will make sure the deployment also fails if there is at least one failure.

docs/architecture.md
vasu2856: yeah. updated color to black to make the text more readable.

docs/architecture.md
vasu2856: have fixed it to render with a mix of horizontal and vertical making it more readable.

docs/architecture.md
vasu2856: yup. have made references to domain and project as singular.

docs/architecture.md
vasu2856: good catch. Have updated the overall architecture doc to be more upto date with what's in the code/repo.

docs/catalog-import-export-guide.md
vasu2856: good catch. Have updated the doc to be up to date and added a bold text in the warning to highlight that the resources not present in the bundle will be deleted.

docs/catalog-import-export-guide.md
vasu2856: yeah... tell me about it :) . Re-basing and merging the conflicts was a pain. Still not 100% sure if the conflict resolution were accurate.

docs/catalog-import-export-guide.md
vasu2856: yeah. I understand the confusion. it was intended to be used for only catalog resources. Makes me wonder if we should support this option at all for launch.

docs/catalog-import-export-guide.md
vasu2856: yeah. not for launch. If we see an ask, we can add it later.

docs/catalog-import-export-guide.md
vasu2856: Not sure if i follow. We won't be replicating the physical data. The expectation is that customer would have that setup already.

docs/catalog-import-export-guide.md
vasu2856: Its documented here for transparency, since its part of the bundle we create for the customer.

docs/catalog-import-export-guide.md
vasu2856: yeah. have fixed this with the dry-run change.

docs/catalog-import-export-guide.md
vasu2856: Are you talking about the dry run feature?

docs/catalog-import-export-quick-reference.md
vasu2856: Will delete the quick reference.

docs/catalog-import-export-guide.md
vasu2856: Decided to remove this option all together.

docs/catalog-import-export-guide.md
vasu2856: yup. Have removed it from the doc.

docs/catalog-import-export-guide.md
vasu2856: the error message lets the customer know what's missing. For eg the error will say the Form Type <name> is not enabled. More on this as part pf dry run PR.

docs/catalog-import-export-guide.md
vasu2856: thanks for catching this. Have reconciled it.

examples/catalog-import-export/app_tests/test_catalog_deployed.py
vasu2856: But manifest doesn't list all the glossary terms rite?

examples/catalog-import-export/app_tests/test_catalog_deployed.py
vasu2856: yup. Fixed it.

examples/catalog-import-export/app_tests/test_catalog_deployed.py
vasu2856: fixed.

examples/catalog-import-export/app_tests/test_catalog_deployed.py
vasu2856: fixed.

examples/catalog-import-export/manifest.yaml
vasu2856: yup. updated the manifest.

examples/catalog-import-export/manifest.yaml
vasu2856: for create? have removed it. We should tackle remove it from schema and code separately as Kiro generated manifests seem to default to this.

examples/catalog-import-export/manifest.yaml
vasu2856: yeah. May be due to merge conflict, i do see them. But have removed it from this manifest.

examples/catalog-import-export/README.md
vasu2856: this is removed along with the update after support

.github/workflows/catalog-import-export.yml
vasu2856: Yes. We have a separate internal module to populate catalog samples. Removing the seed_catalog_data.py

src/smus_cicd/application/application_manifest.py
vasu2856: yup. removed.

src/smus_cicd/helpers/catalog_import.py
vasu2856: removed update after flag to avoid this confusion :)

src/smus_cicd/cli.py
vasu2856: removed it instead.

src/smus_cicd/cli.py
vasu2856: removed it.

tests/integration/catalog-import-export/test_catalog_edge_cases.py
vasu2856: updated the test.

tests/integration/catalog-import-export/test_catalog_edge_cases.py
vasu2856: done

.kiro/specs/datazone-catalog-import-export/design.md
Shnekit: This was related to deletion logic for anything not in the bundle. Since we removed it, it does not apply anymore

.kiro/specs/datazone-catalog-import-export/requirements.md
Shnekit: Cool, will take a look at this in the dry-run PR

.github/workflows/catalog-import-export.yml
Shnekit: Where did we move this to?

docs/catalog-import-export-guide.md
```diff
@@ -0,0 +1,526 @@
...
+| **Data Products** | Data products that bundle one or more data assets for publishing and sharing |
+
+## Key Assumption: Physical Resources Must Have the Same Name
+
+> ⚠️ **IMPORTANT**: Catalog import/export assumes that the underlying physical resources (e.g., Glue Tables, Glue Databases, S3 buckets) referenced by your catalog assets **have the same name in both source and target environments**. The import process matches assets across environments using normalized `externalIdentifier` values or resource names. If the physical resource names differ between environments (e.g., `my-table-dev` vs `my-table-prod`), the matching will fail and resources will be created as new rather than updated. **Additionally, any resources that exist in the target project but are not present in the bundle will be deleted during import.** This means mismatched names can cause the original target resource to be deleted and a duplicate to be created. Ensure your infrastructure provisioning uses consistent resource names across stages, or that your naming conventions allow the normalization logic (which strips AWS account IDs and region strings) to produce matching identifiers.
```
Shnekit, line 30: If the resources not in the bundle deletion logic is with Dry-Run PR, let's make sure we are updating the docs and comments as well, like here

docs/catalog-import-export-guide.md
Shnekit: Yeah, I think I misunderstood this first time I read it. Ignore my comment

docs/catalog-import-export-guide.md
Shnekit: No, as the command for "Verify catalog resources in dev" they can setup automatic verifications and run them with `aws-smus-cicd-cli test`

examples/catalog-import-export/app_tests/test_catalog_deployed.py
Shnekit: Makes sense

Shnekit (approved):
Resolved all the blocking comments. Looks good now. Let's keep track of everything that is supposed to be in dry-run PR related to this. 
Otherwise, LGTM

PR #71, opened 2026-03-06
Fix SMUS compatibility: fallback role resolution and example workflow path alignment

driftnine:
What Changed
Added role resolution fallback logic for SMUS projects when tooling environment role discovery does not return a usable ARN.
Improved project membership handling to correctly derive owner-based role data for SMUS scenarios.
Aligned the example analytics workflow packaging and script path references so deployed workflow tasks resolve ETL scripts consistently.
Validation
Confirmed workflow creation now proceeds with fallback role resolution in SMUS.

### DIFF
--- a/examples/analytic-workflow/dashboard-glue-quick/covid_etl_workflow.yaml
+++ b/examples/analytic-workflow/dashboard-glue-quick/covid_etl_workflow.yaml
@@ -5,7 +5,7 @@ workflow_combined:
       operator: airflow.providers.amazon.aws.operators.glue.GlueJobOperator
       retries: 0
       job_name: setup-covid-db-job
-      script_location: '{proj.connection.default.s3_shared.s3Uri}dashboard-glue-quick/bundle/glue_setup_covid_db.py'
+      script_location: '{proj.connection.default.s3_shared.s3Uri}etl-scripts/bundle/glue_setup_covid_db.py'
       s3_bucket: '{proj.connection.default.s3_shared.bucket}'
       iam_role_name: '{proj.iam_role_name}'
       region_name: '{domain.region}'
@@ -22,7 +22,7 @@ workflow_combined:
       operator: airflow.providers.amazon.aws.operators.glue.GlueJobOperator
       retries: 0
       job_name: summary-glue-job
-      script_location: '{proj.connection.default.s3_shared.s3Uri}dashboard-glue-quick/bundle/glue_covid_summary_job.py'
+      script_location: '{proj.connection.default.s3_shared.s3Uri}etl-scripts/bundle/glue_covid_summary_job.py'
       s3_bucket: '{proj.connection.default.s3_shared.bucket}'
       iam_role_name: '{proj.iam_role_name}'
       region_name: '{domain.region}'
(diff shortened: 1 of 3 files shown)


examples/analytic-workflow/dashboard-glue-quick/manifest.yaml
```diff
@@ -27,18 +27,17 @@ stages:
   dev:
     stage: DEV
     domain:
```
vasu2856, line 29: the domain name will be specific to individual use cases and hence our examples, which run as git hub actions, utilize tag based lookup. I would prefer to continue with the tag based lookup for the examples in the repo.

examples/analytic-workflow/dashboard-glue-quick/covid_etl_workflow.yaml
```diff
@@ -5,7 +5,7 @@ workflow_combined:
       operator: airflow.providers.amazon.aws.operators.glue.GlueJobOperator
       retries: 0
       job_name: setup-covid-db-job
-      script_location: '{proj.connection.default.s3_shared.s3Uri}dashboard-glue-quick/bundle/glue_setup_covid_db.py'
+      script_location: '{proj.connection.default.s3_shared.s3Uri}etl-scripts/bundle/glue_setup_covid_db.py'
```
vasu2856, line 8: Curious, why are we changing the bundle path name?

src/smus_cicd/helpers/datazone.py
```diff
@@ -157,6 +159,28 @@ def get_project_user_role_arn(project_name: str, domain_name: str, region: str)
                         typer.echo(f"✅ DEBUG: Found userRoleArn={role_arn}")
                         return role_arn
 
+        # Fallback: Get role from project owners (for SMUS compatibility)
```
vasu2856, line 162: can't we use [this](https://github.com/aws/CICD-for-SageMakerUnifiedStudio/blob/main/docs/substitutions-and-variables.md#project-properties) in the manifest instead? Why do we need this fallback in the code?

examples/analytic-workflow/dashboard-glue-quick/covid_etl_workflow.yaml
driftnine: the script already lies in dashboard-glue-quick folder in the repo, manifest file pointing to this folder duplicates the folder name

src/smus_cicd/helpers/datazone.py
driftnine: i just tried to fix  in a less interfering way.Ideally it should be the project_role only.

examples/analytic-workflow/dashboard-glue-quick/manifest.yaml
driftnine: sorry thats not an intended change. I forgot to revert my specific use case changes before committing.

Shnekit:
Converting the PR to draft. Feel free to resolve conflicts and raise it again, if you still think the changes are necessary. We did quite a few bug fixes since the time the PR was published.

PR #78, opened 2026-03-12
fix: bug bash fixes — domain resolution, bundle copytree, env var defaults, docs

Shnekit:
## Summary

A collection of fixes found during the SA bug bash session.

## Bug Fixes

**Domain resolution (`utils.py`)**
- Refactored `_resolve_domain_id` to always fall through to `resolve_domain_id` even when no `id`/`name`/`tags` are provided, enabling auto-detection when only one DataZone domain exists in the account.
- Priority order: CloudFormation export → direct `domain.id` → resolve by name/tags/auto-detect.

**DataZone environment lookup (`datazone.py`)**
- Removed the `"tooling"` name filter in `get_project_user_role_arn`. Not all projects name their environment with "tooling", so the filter was preventing role ARN lookup. Now returns the first environment that has a `userRoleArn` provisioned resource.

**Bundle directory copy (`bundle.py`)**
- Fixed `shutil.copytree` to use `dirs_exist_ok=True` and copy contents directly into `target_dir` instead of creating a subdirectory. The old behavior caused a double-nesting bug (`notebooks/notebooks/`) when the include pattern directory name matched the storage item `name`. The zip structure (`{name}/files`) is preserved correctly for deploy.

**Remove env var defaults from manifests**
- Stripped `${VAR:default-region}` defaults from all manifests and examples. The defaults were causing misleading errors (wrong region, project/domain not found) when users forgot to export the variable. Now fails fast with a clear missing-variable error.
- Affected: `dashboard-glue-quick`, `data-notebooks`, `genai`, `ml/training`, `ml/deployment`, `DemoMarketingBundle.yaml`, `quicksight-bootstrap-example.yaml`, MWAA/serverless examples, `README.md`.

**MLflow connection variable cleanup (`ml/training`, `ml/deployment` manifests)**
- Replaced `${STS_REGION}` / `${STS_ACCOUNT_ID}` with `${TEST_DOMAIN_REGION}` / `${AWS_ACCOUNT_ID}` for consistency.

## Documentation

**Quickstart guide**
- Added venv setup steps (create + activate on macOS/Linux/Windows).
- Added Step 3 for setting up a SageMaker Unified Studio domain and project before deploying.
- Clarified that `domain` identifier can be omitted entirely if only one domain exists in the account.
- Fixed broken links: `bundle-manifest.md` → `manifest.md` across all docs.

**Dashboard Glue Quick README**
- Expanded setup steps: env var export, `setup_test_dashboard.py` step, and initial bundle step before deploy. Renumbered steps accordingly.

**`setup_test_dashboard.py`**
- Always resolve account ID via STS instead of optionally falling back to env var.

## Manifest cleanup

- Removed unused `dev` stage from `data-notebooks` and `genai` manifests (not used by integration tests).
- Restored `dev` stage (including `S3_PREFIX` env var) in `ml/training` and `ml/deployment` — required by integration tests for bundling.

## Tests

- Added `TestResolveDomainIdAutoDetect` unit tests: auto-detect single domain, multiple domains error, CloudFormation priority, direct ID short-circuit.
- Added `test_bundle_directory_copy_contents_flat` unit test verifying the copytree fix.

### DIFF
--- a/docs/cli-commands.md
+++ b/docs/cli-commands.md
@@ -964,7 +964,7 @@ All commands support:
 
 ## Configuration Files
 
-### Bundle Manifest
+### Manifest
 - Default location: `manifest.yaml` (current directory)
 - Override with `--manifest` option
 - See [Manifest Reference](manifest.md) for format
--- a/docs/cli-commands.md
+++ b/docs/cli-commands.md
@@ -967,7 +967,7 @@ All commands support:
 ### Bundle Manifest
 - Default location: `manifest.yaml` (current directory)
 - Override with `--manifest` option
-- See [Bundle Manifest Reference](bundle-manifest.md) for format
+- See [Manifest Reference](manifest.md) for format
 - **Error handling**: CLI will error if the default file doesn't exist and no alternative is specified
 
 ### AWS Configuration
--- a/src/smus_cicd/commands/bundle.py
+++ b/src/smus_cicd/commands/bundle.py
@@ -256,13 +256,11 @@ def bundle_command(
                             files_added += 1
                             typer.echo(f"    Copied file: {pattern}")
                         elif os.path.isdir(source_path):
-                            # Copy directory recursively
-                            dest_path = os.path.join(
-                                target_dir, os.path.basename(pattern)
-                            )
+                            # Copy directory contents directly into target_dir
                             shutil.copytree(
                                 source_path,
-                                dest_path,
+                                target_dir,
+                                dirs_exist_ok=True,
                                 ignore=shutil.ignore_patterns(
                                     "*.pyc",
                                     "__pycache__",
@@ -271,7 +269,7 @@ def bundle_command(
                                 ),
                             )
                             # Count files
-                            for root, dirs, files in os.walk(dest_path):
+                            for root, dirs, files in os.walk(target_dir):
                                 files_added += len(files)
                             typer.echo(f"    Copied directory: {pattern}")
                         else:
--- a/src/smus_cicd/helpers/datazone.py
+++ b/src/smus_cicd/helpers/datazone.py
@@ -149,9 +149,9 @@ def get_project_user_role_arn(project_name: str, domain_name: str, region: str) 
         env_names = [env.get("name") for env in environments_response.get("items", [])]
         typer.echo(f"🔍 DEBUG: Found environments: {env_names}")
 
-        # Find tooling environment
+        # Find an environment with UserRoleArn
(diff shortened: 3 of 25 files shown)


docs/cli-commands.md
```diff
@@ -967,7 +967,7 @@ All commands support:
 ### Bundle Manifest
```
vasu2856: We should remove bundle from here as well

src/smus_cicd/commands/bundle.py
```diff
@@ -256,13 +256,11 @@ def bundle_command(
                             files_added += 1
                             typer.echo(f"    Copied file: {pattern}")
                         elif os.path.isdir(source_path):
-                            # Copy directory recursively
-                            dest_path = os.path.join(
```
vasu2856, line 260: Doesn’t this code help create subdirectories within bundle directory , for eg workflows ?

src/smus_cicd/helpers/datazone.py
```diff
@@ -151,19 +151,18 @@ def get_project_user_role_arn(project_name: str, domain_name: str, region: str)
 
         # Find tooling environment
```
vasu2856: Should we remove this comment ?

src/smus_cicd/commands/bundle.py
Shnekit: We are already creating a subdirectory on lines 243 and 244. It is a bit confusing, so let me explain.

Suppose you have a manifest file, which specifies the following:
```
content:
  storage:
    - name: workflows
      include:
        - 'workflows-dir'
```
Here, `workflows` is the name of the entity that we ask the bundle to copy. `workflows-dir` is a local directory from where bundle command is supposed to copy from.

So, the current code will create a bundle structured like this
```
📦 Bundle Contents:
==================================================
├── workflows/
│   └── workflows-dir/
│       ├── customer_churn_prediction.ipynb
│       ...
```
which is double-nesting. And if you write `workflows-dir/` with the trailing slash, then it will fail and does not produce the bundle at all because `os.path.basename("workflows-dir/")` returns an empty string.

But with this fix we are doing the following:
```
📦 Bundle Contents:
==================================================
├── workflows/
│   ├── customer_churn_prediction.ipynb
│    ...
```

vasu2856: got it, thanks! Didn't realize we were replicating the local folder structure.

vasu2856 (approved):
LGTM

PR #83, opened 2026-03-16
debug: add resolver logging and fix MLflow tracking server name in workflows

Shnekit:
### Fix: MLflow tracking server name in GitHub workflows
- `smus-bundle-deploy.yml` and `smus-direct-deploy.yml`: replace `steps.mlflow-setup*.outputs.tracking-server-name` with `format('smus-integration-mlflow-{0}', vars.DOMAIN_REGION)` — the step output was causing errors in the Training GitHub workflow

### Debug: ContextResolver variable resolution logging
Investigating why `{proj.connection.default.s3_shared.s3Uri}` in `ml_deployment_workflow.yaml` is not being resolved correctly during prod ML deployment, resulting in the test account S3 URI being written to the workflow YAML instead of the prod account URI.

Added debug logging to `ContextResolver.resolve()` that prints:
- The variable being resolved and available connection names (so we can see what DataZone actually returns at runtime)
- Which connection matched (or didn't match)
- Which property was found (or missing)

This logging will be removed once the root cause is confirmed.


vasu2856 (commented):
LGTM

PR #101, opened 2026-03-22
fix: correct opensearch client package name to opensearch-py

x4v13r64:
The OpenSearch Python client pip package is named `opensearch-py`, not `opensearchpy`. The importable module is `opensearchpy`, but `pip install opensearchpy` resolves the wrong package name — `opensearch-py` is the correct install target.

This PR corrects the package name in `requirements.txt`.


Shnekit (approved):
LGTM, thanks for contributing :)

PR #128, opened 2026-03-31
feat: destroy command — requirements and design spec

Shnekit:
## Overview

This PR adds the requirements and design spec for the new `destroy` command. No implementation code yet — for team review before implementation begins.

## What's included

- `.kiro/specs/destroy-command/requirements.md` — 12 requirements covering the full feature scope
- `.kiro/specs/destroy-command/design.md` — technical design with architecture, data models, correctness properties, and testing strategy

## Feature summary

The `destroy` command is the inverse of `deploy`. It reads a manifest file and deletes all resources created by `deploy`, in a safe dependency order:

1. Stop active Airflow workflow runs (re-checked at destruction time)
2. Delete workflow-created resources (Glue jobs) via operator registry
3. Delete Airflow serverless workflows
4. Delete QuickSight dashboards, datasets, data sources (by prefix scan)
5. Delete S3 objects at declared `targetDirectory` prefixes
6. Delete DataZone project (only if `project.create: true`)

## Key design decisions

- **Two-phase model**: full validation across all stages before any destructive action
- **Collect-all-errors**: validation never aborts early — all issues reported together before aborting
- **Operator registry**: workflow-created resources (Glue jobs) discovered by parsing workflow YAML and matching operator class names against a registry dict — extensible without if/elif chains
- **Prefix-based QuickSight enumeration**: scans by `Resource_Prefix` rather than reconstructing IDs
- **No `--async` mode**: dependency ordering requires synchronous execution
- **Collision detection**: QuickSight prefix scan and Airflow workflow name lookup both abort if more resources found than declared — even with `--force`

## Testing plan

- Unit tests covering all correctness properties (mocked AWS calls)
- Integration test: extend existing `test_dashboard_glue_quick_workflow.py` with destroy + verify steps after the existing deploy + test steps

## Review focus

Please review requirements and design for correctness, completeness, and any edge cases missed before implementation begins.


.kiro/specs/destroy-command/requirements.md
```diff
@@ -0,0 +1,274 @@
+# Requirements Document
+
+## Introduction
+
+The `destroy` command is the inverse of the `deploy` command for the SMUS CI/CD CLI tool. It reads a manifest file, identifies the target stage(s), and deletes all resources that were directly created by `deploy` — including QuickSight dashboards/datasets/data sources, S3 objects at declared target paths, Airflow serverless workflows, DataZone connections created by bootstrap actions, and optionally the DataZone project. It respects a strict dependency ordering to ensure safe teardown, collects all user confirmations before any destructive action begins, and is idempotent: resources already absent are logged as warnings rather than errors.
+
+The `destroy` command is distinct from the existing `delete` command. The `delete` command only removes the DataZone project and its environments. The `destroy` command removes all deployed content resources (QuickSight, S3, Airflow workflows, workflow-created resources) and conditionally removes the DataZone project.
```
vasu2856: Why do we still have/need delete command?

.kiro/specs/destroy-command/requirements.md
```diff
@@ -0,0 +1,274 @@
...
+
+The `destroy` command is always synchronous. Each deletion step must complete before the next begins, preserving the strict dependency ordering that makes teardown safe. There is no `--async` mode.
```
vasu2856: We are confusing between sequential and synchronous. async runs can be sequential. I don't think we should talk about async, unless we want to support it explicitly. Otherwise, the expectation is same as any other command, which runs sync.

.kiro/specs/destroy-command/requirements.md
```diff
@@ -0,0 +1,274 @@
...
+
+For resources created by Airflow workflow runs (e.g. Glue jobs), destroy uses an operator registry approach: it parses the workflow YAML files declared in `content.workflows`, identifies tasks whose operator type is in the supported registry, extracts the resource name from a well-known static field, and deletes those resources directly via boto3. Resources created by operators not in the registry (e.g. `SageMakerNotebookOperator`, which runs notebooks but creates no persistent named resource itself) are skipped and noted in the output. Resources created dynamically inside notebook code (e.g. Bedrock Agents) remain out of scope.
```
vasu2856, line 7: Do we need this? Would we know whether the resource was created by workflow vs an existing resource which was run as part of workflow? For eg, a glue job operator will run the job it exists and create if it doesn't. 

Also, the workflow itself might not be created for CI/CD. Instead, its solving a business purpose.

.kiro/specs/destroy-command/requirements.md
```diff
@@ -0,0 +1,274 @@
...
+## Requirements
+
+### Requirement 1: CLI Interface
+
+**User Story:** As a DevOps engineer, I want a `destroy` CLI subcommand with consistent options, so that I can tear down deployed resources using the same conventions as other commands.
+
+#### Acceptance Criteria
+
+1. THE Destroy_Command SHALL accept a `--manifest` / `-m` option specifying the path to the manifest file, defaulting to `manifest.yaml`.
+2. THE Destroy_Command SHALL accept a `--targets` / `-t` option specifying a comma-separated list of stage names to destroy, defaulting to all stages defined in the manifest.
```
vasu2856: How can we default to all stages in the manifest? We should make this a required field to specify which target to destroy.

.kiro/specs/destroy-command/requirements.md
```diff
@@ -0,0 +1,274 @@
...
+
+---
+
+### Requirement 3: Pre-Destruction Validation and Confirmation
+
+**User Story:** As a DevOps engineer, I want all validations and confirmations to happen before any resource is deleted, so that I can review the full impact and abort safely.
+
+#### Acceptance Criteria
+
+1. THE Destroy_Command SHALL run a full validation cycle across ALL targeted stages before initiating any destructive action or prompting for confirmation. Resource discovery includes live AWS API calls to resolve S3 connection details, enumerate QuickSight resources by prefix, and check for active Airflow workflow runs.
```
vasu2856, line 77: Will the validations include missing permission errors?

.kiro/specs/destroy-command/requirements.md
```diff
@@ -0,0 +1,274 @@
...
+4. WHEN the validation cycle completes with no errors, THE Destroy_Command SHALL print a full destruction plan listing every resource that will be deleted, grouped by stage and resource type, including resource IDs and S3 prefixes.
+5. WHEN `--force` is not set, after printing the destruction plan THE Destroy_Command SHALL prompt the user for a single explicit confirmation before proceeding.
+6. WHEN the user declines the confirmation prompt, THE Destroy_Command SHALL exit without performing any destructive actions.
+7. WHEN `--force` is set, THE Destroy_Command SHALL skip the confirmation prompt and proceed directly to destruction after a clean validation cycle.
+8. WHEN one or more targeted stages contain Airflow_Workflows with active Workflow_Runs AND `--force` is not set, THE Destroy_Command SHALL include the active runs in the destruction plan and ask the user whether to force-stop them or abort, as part of the single confirmation step.
```
vasu2856: isn't force-stopping or aborting a workflow the same thing? 

I realized that the abort here is aborting the command after reading the next one. If we are not already, we should be more explicit about aborting the destroy command vs aborting job runs/workflow runs.

docs/cli-commands.md
```diff
@@ -519,7 +520,43 @@ Domain: cicd-test-domain (us-east-1)
 **What this shows:** The test command runs Python tests from the configured test folder against your deployed bundle. Tests receive environment variables with domain ID, project ID, and other context information to validate the deployment. This ensures your pipeline is working correctly after deployment and provides automated validation of your data workflows.
 
 ### 8. Clean Up Resources
+
+Use `delete` to remove the DataZone project only, or `destroy` to remove all deployed resources (QuickSight, S3, Airflow workflows, Glue jobs, catalog resources, and optionally the project).
```
vasu2856: Please include a warning here to highlight that this is not a reversible operation and they should handle it with care.

vasu2856: Also, left this comment somewhere else, why do we still need delete?

docs/cli-commands.md
```diff
@@ -952,6 +990,160 @@ aws-smus-cicd-cli delete -t test --force -o JSON
 - CloudFormation stacks are deleted automatically when projects are removed
 - Use `--async` for faster execution when managing multiple targets
 
+### 9. destroy - Destroy All Deployed Resources
```
vasu2856: Seems like a repeat of #8 above. What's the difference between these sections?

docs/rollback-guide.md
```diff
@@ -0,0 +1,442 @@
...
+This guide explains how to recover from a bad deployment using the `destroy` and `deploy` commands — both manually and through automated rollback workflows.
+
+---
+
+## Overview
+
+A "bad deployment" can mean different things:
+
+- A workflow fails after deployment (broken Glue job, bad notebook, misconfigured Airflow DAG)
+- A QuickSight dashboard is broken or shows wrong data
```
vasu2856: i would leave wrong data out of this as  data is use case specific. I don't want to give a false sense that we will help deploy 'data'

docs/rollback-guide.md
```diff
@@ -0,0 +1,442 @@
...
+- A deployment introduced a regression in a previously working stage
+- You need to revert to a known-good version of your application
```
vasu2856: these two need to be merged into a single entry.

docs/cli-commands.md
```diff
@@ -952,6 +990,160 @@ aws-smus-cicd-cli delete -t test --force -o JSON
...
+- Resolves domain and project IDs via DataZone
+- Resolves S3 connections to determine bucket/prefix targets
+- Enumerates QuickSight resources by prefix and detects collisions
+- Checks for active Airflow workflow runs
+- Fetches and parses workflow YAML files from S3 to discover workflow-created resources (e.g. Glue jobs)
+- Collects ALL errors across ALL stages before aborting — no early exit
+
+**Phase 2 — Destruction (after confirmation):**
+
+Resources are deleted in this fixed dependency order per stage:
```
vasu2856, line 824: We are not talking about any of the catalog resources.

docs/rollback-guide.md
vasu2856: We should also talk about Catalog being in undesirable state or tests failing

docs/rollback-guide.md
```diff
@@ -0,0 +1,442 @@
...
+
+The SMUS CI/CD CLI supports two recovery strategies:
+
+| Strategy | When to use | How |
+|----------|-------------|-----|
+| **Redeploy previous version** | You have a previous bundle or git state | `deploy` with the previous artifact |
```
vasu2856, line 22: We should highlight the difference between the two. I would callout that this operation will not remove anything added from the previous deployment will only create/update what exists.

docs/rollback-guide.md
```diff
@@ -0,0 +1,442 @@
...
+```bash
+# 1. Identify the previous good bundle in S3
+aws s3 ls s3://your-bundle-bucket/bundles/ --recursive | sort
+
+# 2. Download the previous bundle locally
+aws s3 cp s3://your-bundle-bucket/bundles/MyApp-v1.2.0.zip ./artifacts/MyApp.zip
+
+# 3. Deploy from the local bundle
+#    (deploy picks up ./artifacts/MyApp.zip automatically)
+aws-smus-cicd-cli deploy --stages prod --manifest manifest.yaml
```
vasu2856: shouldn't it be --targets?

docs/rollback-guide.md
```diff
@@ -0,0 +1,442 @@
...
+Redeploying a previous bundle is sufficient when:
+- The issue is in application code (notebooks, Glue scripts, DAGs)
+- S3 files are the primary deployed artifact
+- The DataZone project and connections are still healthy
+- No new AWS resources were created by the bad deployment that need cleaning up
+
+### When This Is NOT Enough
+
+A simple redeploy won't fully restore state if:
+- The bad deployment created new AWS resources (Glue jobs, QuickSight dashboards) that conflict with the previous version
```
vasu2856: won't these get auto corrected with a new deployment? I would highlight scenarios where the jobs were created by mistake and won't be part of updated bundle.

docs/rollback-guide.md
```diff
@@ -0,0 +1,442 @@
...
+- Workflow-created resources (e.g. Glue jobs) have stale configuration from the bad deployment
+- You need a completely clean environment before redeploying
+
+In those cases, use Strategy 2.
+
+---
+
+## Strategy 2: Destroy and Redeploy (Clean Slate)
+
+This is the most reliable rollback approach. It removes all resources deployed by the bad version, then redeploys the known-good version from scratch.
```
vasu2856: I'm not sure if we should say 'reliable'. May be complete? We might have some resources which the destroy might not remove. So I would avoid sweeping statements here.

docs/rollback-guide.md
```diff
@@ -0,0 +1,442 @@
...
+2. Print a full destruction plan
+3. Stop any active Airflow workflow runs
+4. Delete workflow-created resources (Glue jobs, etc.)
+5. Delete Airflow workflows
+6. Delete QuickSight dashboards, datasets, and data sources
+7. Delete S3 objects at declared target paths
+8. Delete catalog resources (if configured)
+9. Delete the DataZone project (only if `project.create: true`)
+
+> **Note:** If `project.create: false` in your manifest, the DataZone project is preserved. Only deployed content is removed. This is the typical case for rollbacks — you keep the project and just clean up the deployed resources.
```
vasu2856, line 94: what if the project create is true?

docs/rollback-guide.md
```diff
@@ -0,0 +1,442 @@
...
+
+### Step 2: Redeploy the Previous Good Version
+
+```bash
+# Option A: Redeploy from a previous bundle artifact
+#   Download the specific versioned bundle, then deploy
+aws s3 cp s3://your-bundle-bucket/bundles/MyApp-v1.2.0.zip ./artifacts/MyApp.zip
+aws-smus-cicd-cli deploy --stages prod --manifest manifest.yaml
```
vasu2856: We should fix everywhere we have --stages to --targets

docs/rollback-guide.md
```diff
@@ -0,0 +1,442 @@
...
+            }
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+```
+
+---
+
+## Bundle-Based Rollback (Recommended for Production)
+
+The most reliable rollback strategy is bundle-based deployment, where each deployment creates a versioned artifact. This gives you instant rollback to any previous version without needing to re-run the build.
```
vasu2856: 'most reliable' As opposed to? I would update the docs to not say reliable and instead recommended strategy.

src/smus_cicd/commands/destroy.py
```diff
@@ -0,0 +1,1825 @@
...
+
+from ..application import ApplicationManifest
+from ..helpers import s3 as s3_helper
+from ..helpers.airflow_serverless import (
+    delete_workflow,
+    generate_workflow_name,
+    is_workflow_run_active,
+    list_workflow_runs,
+    list_workflows,
+    stop_workflow_run,
```
vasu2856: why do we need to stop workflow run? when we delete workflow, it automatically stops the workflow run.

src/smus_cicd/commands/destroy.py
```diff
@@ -0,0 +1,1825 @@
+"""Destroy command for SMUS CI/CD CLI.
+
+Implements the inverse of deploy: reads a manifest, discovers all deployed
```
vasu2856, line 3: We need a mechanism to make sure this doesn't drift from the deploy implementation.

src/smus_cicd/commands/destroy.py
```diff
@@ -0,0 +1,1825 @@
...
+        if bucket:
+            raw_targets.append(
+                S3Target(
+                    bucket=bucket,
+                    prefix=full_prefix,
+                    connection_name=conn_name,
+                )
+            )
+
+    # Process git entries
```
vasu2856: git?

src/smus_cicd/commands/destroy.py
```diff
@@ -0,0 +1,1825 @@
...
+            resources.append(
+                ResourceToDelete(
+                    resource_type="airflow_workflow",
+                    resource_id=workflow_arn,
+                    stage=stage_name,
+                    metadata={"workflow_name": workflow_name},
+                )
+            )
+
+            # Check for active runs
```
vasu2856: we don't need these. Just delete the workflow and let MWAA take care of the runs.

src/smus_cicd/commands/destroy.py
```diff
@@ -0,0 +1,1825 @@
...
+                    f"[{stage_name}] Could not list runs for workflow "
+                    f"'{workflow_name}': {e}"
+                )
+        else:
+            warnings.append(
+                f"[{stage_name}] Workflow '{workflow_name}' not found — "
+                "will be skipped during destruction"
+            )
+
+        # Fetch and parse workflow YAML from S3 to discover created resources.
```
vasu2856: i don't think we should rely on the yaml file in S3 for source. Instead, get the workflow source from MWAA serverless APIs. But on a higher level note, i don't think we should parse the workflow to figure out if we have created any resources. I would instead just delete the workflow and leave it at that.

src/smus_cicd/commands/destroy.py
```diff
@@ -0,0 +1,1825 @@
...
+                from ..helpers.catalog_import import (
+                    _is_managed_resource,
+                    _search_target_resources,
+                    _search_target_type_resources,
+                )
+
+                dz_client = boto3.client("datazone", region_name=effective_region)
+
+                # Glossaries
+                for item in _search_target_resources(
```
vasu2856: shouldn't these be driven by what's in the manifest instead of searching for them via DZ APIs?

src/smus_cicd/commands/destroy.py
```diff
@@ -0,0 +1,1825 @@
...
+                warnings.append(
+                    f"[{stage_name}] Could not enumerate catalog resources: {e}"
+                )
+        elif catalog_disabled:
+            warnings.append(
+                f"[{stage_name}] Catalog deletion is disabled (disable: true) — skipping."
+            )
+
+    # --- Bootstrap connections ---
+    # Only discover if project is NOT being deleted (project deletion cascades to connections)
```
vasu2856: what does it mean to delete project? Deleting a project would drop catalog resources and connections. So we wouldn't need to drop these one by one.

src/smus_cicd/commands/destroy.py
```diff
@@ -0,0 +1,1825 @@
...
+    for qs_type in (
+        "quicksight_dashboard",
+        "quicksight_dataset",
+        "quicksight_data_source",
+    ):
+        for resource in validation_result.resources:
+            if resource.resource_type != qs_type:
+                continue
+
+            resource_id = resource.resource_id
```
vasu2856: shouldn't the deletion happen in order? dashboard->datasource->dataset?

src/smus_cicd/commands/destroy.py
```diff
@@ -0,0 +1,1825 @@
...
+                    ResourceResult(
+                        resource_type="s3_prefix",
+                        resource_id=resource.resource_id,
+                        status="not_found",
+                        message="S3 prefix is empty or does not exist",
+                    )
+                )
+                continue
+
+            s3_helper.delete_objects(bucket, object_keys, region=effective_region)
```
vasu2856: Will this delete the bundle as well?

src/smus_cicd/commands/destroy.py
```diff
@@ -0,0 +1,1825 @@
...
+                    resource_type="stage",
+                    resource_id=stage_name,
+                    status="error",
+                    message=str(e),
+                )
+            ]
+        all_results[stage_name] = stage_results
+
+    # -----------------------------------------------------------------------
+    # 9. Summary
```
vasu2856, line 237: This file is too long. I would split into helper, validator and leave destroy to primarily take care of the operation.

src/smus_cicd/helpers/operator_registry.py
```diff
@@ -0,0 +1,51 @@
+"""
+Operator registry for the destroy command.
```
vasu2856, line 2: Not sure if we need this.

src/smus_cicd/helpers/quicksight.py
```diff
@@ -707,6 +707,94 @@ def trigger_dataset_ingestion(
         )
 
 
+def list_dashboards(
```
vasu2856, line 710: Why do we need to list the dashboards? can't we just directly fetch the ones mentioned in the manifest?

.kiro/specs/destroy-command/requirements.md
Shnekit: Document that Glue jobs are being deleted if they are mentioned in the workflow def

.kiro/specs/destroy-command/requirements.md
Shnekit: See, if we can add them. Otherwise, we can run destroy multiple times

.kiro/specs/destroy-command/requirements.md
Shnekit: Abort the destroy command

docs/cli-commands.md
Shnekit: Will rerun the doc updates

docs/rollback-guide.md
Shnekit: yes

.kiro/specs/destroy-command/requirements.md
Shnekit: Deleting in the next rev

.kiro/specs/destroy-command/requirements.md
Shnekit: Fixing in the next PR

.kiro/specs/destroy-command/requirements.md
Shnekit: Kept the logic for deleting resources under each workflow. Also added the warning to cli-commands docs and destroy command confirmation that reads:
> ⚠️  WARNING: The destroy command deletes ALL resources listed "
        "above based on the manifest and workflow definitions. This includes "
        "resources that may have been created manually by users, not only those "
        "created by the CI/CD tool. Review the destruction plan carefully before "
        "confirming.

.kiro/specs/destroy-command/requirements.md
Shnekit: Made the flag required, so that there is no way to run this for all stages accidentally

.kiro/specs/destroy-command/requirements.md
Shnekit: I think I am gonna wait for your dry-run CR to be out and see, if I can use your existing IAM validation logic to destroy resources.

docs/cli-commands.md
Shnekit: Added

Shnekit: Removed delete functionality

docs/cli-commands.md
Shnekit: Refactored the whole README, so it should be good now

docs/cli-commands.md
Shnekit: I think the resources are actually under number 7, or are you talking about something else?

docs/rollback-guide.md
Shnekit: I think I covered this in the sections When It is Enough and When It is NOT Enough, but lmk, if you want more details anywhere

docs/rollback-guide.md
Shnekit: Yeah, that's exactly what I am saying here. I can write:
> The bad deployment created new AWS resources (Glue jobs, QuickSight dashboards) that will not be part of the updated bundle

docs/rollback-guide.md
Shnekit: Should I remove this project.create stuff? Otherwise, we can take care of everything all at once when getting rid of this feature entirely.

src/smus_cicd/commands/destroy.py
Shnekit: Removed

src/smus_cicd/commands/destroy.py
Shnekit: Added 2 resource_types lists for deploy and destroy with a comment to update both when drafting a new deployment resource.

Let me know, if that is good enough. We might be able to create a function which will query manifest for new fields not included in the destroy functionality, but not sure if this method is deterministic either

src/smus_cicd/commands/destroy.py
Shnekit: There are resources in S3 which can be sourced from git  per the deploy functionality. This is what this block is for

src/smus_cicd/commands/destroy.py
Shnekit: Removed

src/smus_cicd/commands/destroy.py
Shnekit: Fixed.

Keeping the reading the destroying stuff created by WF logic for now, because it is better for rollback and fresh-start use cases that we want to use it for.

Will wait for customer feedback whether they need to remove that functionality

src/smus_cicd/commands/destroy.py
Shnekit: Does manifest have everything we are deploying for catalog? I think deploy is referencing catalog_export.json for deploying the exact resources, for which we need a bundle to destroy properly. So, decided to deliberately destroy all catalog resources we find and documented this in the READMEs everywhere.

Do you think it is okay? Or should we read the bundle (also not too reliable)

src/smus_cicd/commands/destroy.py
Shnekit: Googled it, dashboards seem to depend on datasets, and datasets depend on data sources. So, I think it should be dashboards -> datasets -> data sources

src/smus_cicd/commands/destroy.py
Shnekit: This only deletes the deploy S3 target resources. I think bundle is stored somewhere else, so it is not deleted

src/smus_cicd/commands/destroy.py
Shnekit: Yeah, I could have skipped the catalog deletion logic on project deletion as well. But keeping for now, since we are getting rid of project creation as a feature anyway

src/smus_cicd/commands/destroy.py
Shnekit: Googled this. I think dashboards depend on datasets and datasets depend on datasources, so dashboards -> datasets -> datasources should be the correct deletion order. 

At least it works on our QuickSight example :)

src/smus_cicd/commands/destroy.py
Shnekit: This is deleting everything that deploy command creates. I believe bundle is saved somewhere outside of this path, so it will stay

src/smus_cicd/commands/destroy.py
Shnekit: Fixed

src/smus_cicd/helpers/operator_registry.py
Shnekit: Keeping for now, can remove this later, if customers do not want it

src/smus_cicd/helpers/quicksight.py
Shnekit: In the manifest we are only listing the dashboard name. Any account can have any number of identical dashboard names and they can also be changed after creation. So, doing a prefix search is a little safer, though still not entirely deterministic

docs/cli-commands.md
vasu2856: from what i remember, the checklist below was missing catalog resources earlier. But i see it now.

docs/rollback-guide.md
vasu2856: yeah. works

src/smus_cicd/commands/destroy.py
vasu2856: Why couldn't we use a shared list in the code? All we need to see is if we add a new type to deploy, the tests for destroy fails highlighting that we need to register how to destroy that resource.

src/smus_cicd/helpers/quicksight.py
vasu2856: not sure if i follow the reasoning. We should use the same logic to identify the resource as the deploy. If deploy looks up using name, destroy should do the same as well.

Shnekit: Deploy command is actually using dashboard_id look up as well. Dashboard name is used as the name of the exported .qs file (which QuickSight generates itself I think).

src/smus_cicd/commands/destroy.py
Shnekit: There is no concrete map between the supported resources and the actual implementation in deploy or destroy. If you have just 1 shared list, how can it tell that the new code added is actually a new supported feature and it should be added to the list and destroy logic should be created?

vasu2856 (approved):
LGTM

PR #139, opened 2026-04-02
Feature/catalog import export - Remove resource deletion as part of Deploy

vasu2856:
- Update design documentation to clarify that target resources not in bundle are logged but never deleted
- Remove references to delete operations and update import summary to use "skipped" instead of "deleted"
- Rename internal helper from `_identify_resources_to_delete` to `_identify_extra_target_resources` to reflect logging-only behavior
- Update function docstrings and return value documentation across design, requirements, and testing guide
- Revise example README and deployment guide to reflect non-destructive import behavior
- Update unit tests to verify logging of extra resources instead of deletion operations
- Clarify in deploy command and catalog import helper that import is additive and preserves target resources


.kiro/specs/datazone-catalog-import-export/design.md
```diff
@@ -2,7 +2,7 @@
...
...r-side ownership filtering, while the SearchTypes API requires client-side filtering by `owningProjectId` on response items (it does not support the `owningProjectIdentifier` parameter). For assets, an additional `get_asset` API call per item enriches search results with full details including `formsOutput`, since the Search API only returns summary data. The export captures each asset's and data product's `listingStatus` to preserve the source publish state. The manifest configuration is intentionally simple: only `enabled` (boolean), `skipPublish` (boolean), and `assets.access` (array) — no filter options of any kind exist in the manifest. During deployment, a new `CatalogImporter` component reads the exported JSON, builds an identifier mapping between source and target projects using externalIdentifier (with normalization) or name as fallback, creates or updates resources in dependency order via DataZone create/update APIs, and publishes assets and data products that were published (listingStatus == "ACTIVE") in the source project unless `skipPublish` is set to true. Resources that exist in the target project but are not in the bundle are logged for visibility but never deleted.
```
Shnekit, line 5: Hmmm, I guess it is nice to log them, but do we even need that? How much complexity is it adding?

Shnekit: Nvm, I guess it is useful for resources previously created by CI/CD, so that the customer knows what is still remaining. All good

docs/catalog-import-export-guide.md
```diff
@@ -27,7 +27,7 @@ The catalog import/export feature supports the following DataZone resource types
...
+> ⚠️ **IMPORTANT**: Catalog import/export assumes that the underlying physical resources (e.g., Glue Tables, Glue Databases, S3 buckets) referenced by your catalog assets **have the same name in both source and target environments**. The import process matches assets across environments using normalized `externalIdentifier` values or resource names. If the physical resource names differ between environments (e.g., `my-table-dev` vs `my-table-prod`), the matching will fail and resources will be created as new rather than updated. Ensure your infrastructure provisioning uses consistent resource names across stages, or that your naming conventions allow the normalization logic (which strips AWS account IDs and region strings) to produce matching identifiers.
```
Shnekit, line 30: I would include the warning that CI/CD is not deleting resources that were once included in the previous bundle versions. As a customer, I would have probably expected some CFN-like behavior here, which we are not providing

Shnekit: I know we are saying everywhere that resources not in the bundle are not deleted, but I think it is worth to point out that it also means that it does not touch resources created by previous deployments, not just stuff created manually

src/smus_cicd/helpers/catalog_import.py
```diff
@@ -48,8 +49,8 @@
     "dataProducts",
 ]
 
-# Deletion order (reverse): DataProducts → Assets → AssetTypes → FormTypes → GlossaryTerms → Glossaries
-DELETION_ORDER = list(reversed(CREATION_ORDER))
+# Reverse dependency order: used for reporting extra resources in target
```
Shnekit, line 52: Nit: we do not necessarily need that in reverse order anymore, right? :D

Shnekit:
Do we need any integ test refactors?

.kiro/specs/datazone-catalog-import-export/design.md
vasu2856: yeah Exactly. Its going to be useful for customers to know what's left.

Shnekit (dismissed):
Left several minor comments, but nothing blocking IMO.

LGTM

docs/catalog-import-export-guide.md
vasu2856: yup. Have added a new warning and updated some of the existing sections as well to make it explicit that we don't delete the resources created as part of previous deployments as well.

Shnekit (approved):
Looks good!

PR #141, opened 2026-04-02
docs: add Kiro skill for investigating GitHub Actions workflow failures

Shnekit:
## Summary

Adds a reusable Kiro skill that documents the step-by-step procedure for diagnosing CI/CD workflow failures, based on the investigation patterns used during today's debugging session.

## Skill: `.kiro/skills/investigate-workflow-failure.md`

The skill follows a decision tree:

1. **Quick triage** — use `gh run view` to surface errors immediately. If the failure reason is obvious (permission error, missing secret, CLI failure), stop here.

2. **Notebook execution failures** — if `AlgorithmError: [SM-111]` is present, proceed to deeper investigation:
   - Extract the SageMaker execution UUID and injected params from the logs
   - Find the notebook output tar in S3 using the UUID
   - Download, extract, and parse the output notebook for cell-level errors
   - Check injected parameters vs expected

3. **Common root causes reference table** — covers the issues encountered today: region injection, boto_session missing, DataZone SDK_UNKNOWN_MEMBER, deprecated Bedrock models, workflow version quota.

### DIFF
--- a/.kiro/skills/investigate-workflow-failure.md
+++ b/.kiro/skills/investigate-workflow-failure.md
@@ -1,0 +1,126 @@
+# Investigate GitHub Actions Workflow Failure
+
+Use this skill when a GitHub Actions workflow run has failed and you need to diagnose the root cause.
+
+## Step 1: Get the High-Level Failure Summary
+
+```bash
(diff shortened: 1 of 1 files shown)


vasu2856 (approved):
Nice! LGTM!!

PR #154, opened 2026-04-08
feat: idc_domain_support

Shnekit:
Add IdC domain support for data-notebooks example

- Add test-idc stage to data-notebooks manifest targeting smus-idc-test-domain
- Add deploy-idc job to GitHub workflow for parallel IdC domain testing
- Add default.spark bootstrap connection for IdC domains (missing in project-scoped roles)
- Add idc_domain_data_notebooks_setup.py script for VPC networking (S3 endpoint, NAT gateway) and Lake Formation permissions
- Bump notebook workflow retries from 0 to 1 to handle intermittent kernel timeouts
- Fix link in admin-quickstart docs


docs/getting-started/admin-quickstart.md
```diff
@@ -54,7 +54,7 @@ The CLI supports three deployment approaches:
 2. **Bundle-Based** - Create versioned bundles, then deploy them to stages (all content from bundle)
 3. **Hybrid** - Create bundle for static content (data, configs), pull workflows from git branches per stage
 
-**See more:** [Deployment Approaches Guide](../deployment-approaches.md)
+**See more:** [Deployment Approaches Guide](../deployment.md)
```
vasu2856: I don't see this doc anywhere.

examples/analytic-workflow/data-notebooks/workflows/parallel_notebooks_workflow.yaml
```diff
@@ -3,7 +3,7 @@ notebooks_workflow:
   tasks:
     nb_churn:
       operator: airflow.providers.amazon.aws.operators.sagemaker_unified_studio.SageMakerNotebookOperator
-      retries: 0
+      retries: 1
```
vasu2856, line 6: curious why retry had to be bumped up and is 1 enough?

examples/analytic-workflow/data-notebooks/idc_domain_project_setup.py
```diff
@@ -0,0 +1,452 @@
+#!/usr/bin/env python3
+"""
+Setup script for data-notebooks example on IdC-based domains.
```
vasu2856, line 3: The file name is a bit misleading. Isn't what the script is doing more related to project setup? How is this related to Data Notebook?

examples/analytic-workflow/data-notebooks/manifest.yaml
```diff
@@ -48,5 +48,45 @@ stages:
...
         trailLogs: true
+  test-idc:
+    stage: TEST
+    domain:
+      name: smus-idc-test-domain
+      region: ${TEST_DOMAIN_REGION}
+    project:
+      name: test-marketing
+      owners:
+      - Eng1
```
vasu2856: What is this for?

examples/analytic-workflow/data-notebooks/manifest.yaml
```diff
@@ -48,5 +48,45 @@ stages:
...
         workflowName: notebooks_workflow
         trailLogs: true
+  test-idc:
+    stage: TEST
+    domain:
+      name: smus-idc-test-domain
+      region: ${TEST_DOMAIN_REGION}
+    project:
+      name: test-marketing
+      owners:
```
vasu2856, line 58: Do we need the project role here?

docs/getting-started/admin-quickstart.md
Shnekit: Oh I think I confused it with https://github.com/aws/CICD-for-SageMakerUnifiedStudio/blob/main/docs/development.md. 

Not sure, which doc it was supposed to reference. I will remove the reference altogether

examples/analytic-workflow/data-notebooks/workflows/parallel_notebooks_workflow.yaml
Shnekit: I think I initially saw some flaky timeout issues with this test. I will observe the test for a while and then revert it back to 0 retries, if IdC test behavies

examples/analytic-workflow/data-notebooks/idc_domain_project_setup.py
Shnekit: Yeah sure, I can rename it. This was the setup to make the example work, so named it like this. But I do not have any strong preference in the name

examples/analytic-workflow/data-notebooks/manifest.yaml
Shnekit: Tbh, not sure, this is a carry-over from the other example manifests. If think this can be removed.

examples/analytic-workflow/data-notebooks/manifest.yaml
Shnekit: So, these are just used by deploy to add the roles as project members, so they have DZ API access. Project role already has all the accesses, so it is not needed

examples/analytic-workflow/ml/deployment/workflows/ml_deployment_workflow.yaml
```diff
@@ -3,11 +3,11 @@ ml_deployment_workflow:
   tasks:
     ml_deployment_notebook:
       operator: "airflow.providers.amazon.aws.operators.sagemaker_unified_studio.SageMakerNotebookOperator"
-      retries: 0
+      retries: 1
       input_config:
         input_path: "ml/bundle/deployment-workflows/ml_deployment_notebook.ipynb"
         input_params:
-          model_s3_uri: "{proj.connection.default.s3_shared.s3Uri}ml/bundle/model-artifacts/output/model.tar.gz"
+          model_s3_uri: "{proj.connection.default.s3_shared.s3Uri}ml/bundle/model-artifacts/latest/output/model.tar.gz"
```
vasu2856, line 10: is the model artifacts versioned now? curious why did we have to make this change?

examples/analytic-workflow/ml/training/idc_domain_project_setup.py
```diff
@@ -0,0 +1,185 @@
...
+        resources = {r["name"]: r["value"] for r in detail.get("provisionedResources", [])}
+        role_arn = resources.get("userRoleArn")
+        if role_arn:
+            print(f"   Role: {role_arn}")
+            return {"project_id": project_id, "role_arn": role_arn, "role_name": role_arn.split("/")[-1]}
+    print("❌ No userRoleArn found")
+    return None
+
+
+def setup_mlflow_permissions(role_name, region, dry_run=False):
```
vasu2856, line 66: Mlflow launched serverless version. We should setup an example for that. This should simplify some of these.

vasu2856 (approved):

examples/analytic-workflow/ml/deployment/workflows/ml_deployment_workflow.yaml
Shnekit: I think it used to use the version of the model uploaded manually before... Now, we are using the version which Is copied by the deploy command from the bundle. Also, confirmed it works for both IdC and IAM domains

examples/analytic-workflow/ml/training/idc_domain_project_setup.py
Shnekit: Noted. We can address this later

PR #162, opened 2026-04-14
refactor(bundle-checker): extract bundle requirement logic into reusable helper

vasu2856:
- Add best-effort validation disclaimer to design.md clarifying that dry run does not guarantee deployment success
- Add best-effort validation disclaimer to requirements.md with detailed explanation of potential failure conditions
- Update requirements.md success message requirement to include disclaimer about best-effort nature
- Update tasks.md to mark dependency checker implementation (task 6) as complete
- Update tasks.md to mark DryRunEngine orchestrator implementation (task 8) as complete
- Ensures users understand dry run limitations and potential causes of deployment failure despite passing validation
- Extract manifest bundle requirement check into `manifest_requires_bundle()` helper function in bundle_storage.py
- Replace inline bundle detection logic in deploy.py with call to new helper function
- Update BundleChecker to use same helper for consistent bundle requirement evaluation across deploy and dry-run commands
- Add comprehensive docstring explaining when bundles are required (storage with connectionName or git items)
- Add unit tests verifying checker correctly handles both bundle-required and bundle-optional scenarios
- Allows bundle-less workflows (e.g. ML Training) to proceed without errors when no bundle is needed


.kiro/specs/deploy-dry-run/design.md
```diff
@@ -4,6 +4,8 @@
 
 The deploy dry-run feature adds a `--dry-run` flag to the existing `smus-cicd deploy` command. When active, the CLI executes every deployment phase in read-only mode: it loads and validates the manifest, explores the bundle archive, verifies IAM permissions, checks AWS resource reachability, simulates each deployment phase (project init, storage, git, catalog import, QuickSight, workflows, bootstrap actions), validates pre-existing resource dependencies (Glue Data Catalog resources including tables, views, and databases; data sources; form types; asset types), and produces a structured report — all without creating, modifying, or deleting any resources.
 
+**Important: Best-Effort Validation.** The dry run is a best-effort validation mechanism. A successful dry run significantly reduces the risk of deployment failure but does not guarantee that the actual deployment will succeed. Conditions such as transient AWS service errors, IAM policy changes between validation and deployment, concurrent resource modifications by other actors, eventual consistency delays, and service quotas or throttling limits may cause a deployment to fail even after a clean dry-run report.
```
Shnekit, line 7: I think the biggest contributor here would be operations performed by resources created by workflows, since we perform no validations on them

src/smus_cicd/commands/dry_run/checkers/bundle_checker.py
```diff
@@ -56,7 +56,13 @@ def check(self, context: DryRunContext) -> List[Finding]:
     def _resolve_bundle_path(
         self, context: DryRunContext, findings: List[Finding]
     ) -> str | None:
-        """Resolve the bundle path, falling back to ./artifacts directory."""
+        """Resolve the bundle path, falling back to ./artifacts directory.
+
+        When no bundle is found and the deployment configuration does not
+        require one (no storage items with a connectionName, no git items),
+        the checker reports a WARNING instead of an ERROR so that
```
Shnekit: Do we even need a warning in this case?

Shnekit: Oh, read the code below. Seems like we are still throwing an error, just conditionally now. Ideally, need to fix the comment here then :)

Shnekit (dismissed):
Left 1 small nit, otherwise LGTM

src/smus_cicd/commands/dry_run/checkers/bundle_checker.py
vasu2856: yup. updated the comment.

PR #179, opened 2026-04-21
chore: remove Preview notices and fix duplicate language badges

Shnekit:
- Remove Preview blockquote notices from all READMEs (main, getting-started, examples, tests, and all 6 translations)
- Fix translation script to strip source language badges before translating, preventing duplicate badge panels in translated output

The translation workflow will re-run after merge and produce clean translations without the duplicate badges.

### DIFF
--- a/docs/scripts/translate-readme-chunked.py
+++ b/docs/scripts/translate-readme-chunked.py
@@ -90,7 +90,16 @@ def main():
     print(f"Reading {readme_path}...")
     with open(readme_path, 'r', encoding='utf-8') as f:
         source_content = f.read()
-    
+
+    # Strip the language badge panel from source — the translation script
+    # adds its own badge panel with correct relative paths and highlighting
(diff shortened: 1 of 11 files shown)


docs/scripts/translate-readme-chunked.py
```diff
@@ -90,7 +90,16 @@ def main():
     print(f"Reading {readme_path}...")
     with open(readme_path, 'r', encoding='utf-8') as f:
         source_content = f.read()
-    
+
```
vasu2856, line 93: What's this for?

vasu2856 (approved):
LGTM

docs/scripts/translate-readme-chunked.py
Shnekit: On the languages page, it is creating a duplicate language panel, so getting rid of it here

PR #265, opened 2026-07-01
feat(notebooks): Add spec for native Data Notebooks support

Shnekit:
## Summary

Add requirements, design, and implementation task list for promoting SageMaker Unified Studio notebooks across environments using DataZone Import/Export APIs.

## Key Design Decisions

- **Upsert strategy**: Import creates a new notebook, polls until ACTIVE, applies metadata via UpdateNotebook, then deletes the old version with the same name
- **Name-based matching**: Notebooks are identified by name across environments (no ID-based filters)
- **Unique names enforced**: Bundle command fails if duplicate notebook names are detected
- **Follows existing patterns**: Mirrors catalog import/export architecture

## APIs Used

- ListNotebooks, GetNotebook, StartNotebookExport, GetNotebookExport
- StartNotebookImport, UpdateNotebook, DeleteNotebook

## Spec Files

- `.kiro/specs/data-notebooks-support/requirements.md` — 12 requirements
- `.kiro/specs/data-notebooks-support/design.md` — Architecture, components, 10 correctness properties
- `.kiro/specs/data-notebooks-support/tasks.md` — 24 implementation tasks

## Out of Scope

- Notebook schedules (no public API for schedule management)

### DIFF
--- a/.kiro/specs/data-notebooks-support/design.md
+++ b/.kiro/specs/data-notebooks-support/design.md
@@ -2,25 +2,31 @@
 
 ## Overview
 
-This feature adds native Data Notebooks support to the SMUS CI/CD CLI, enabling promotion of SageMaker Unified Studio notebooks across environments (dev → test → prod) using the DataZone Notebook Import/Export APIs. The implementation follows the existing CLI architecture patterns — specifically mirroring the catalog import/export approach — with two new helper modules (`notebook_export.py` and `notebook_import.py`) and integration into the existing `bundle`, `deploy`, `destroy`, and `dry-run` commands.
+This feature adds native Data Notebooks support to the SMUS CI/CD CLI, enabling promotion of SageMaker Unified Studio notebooks across environments (dev → test → prod) using the DataZone Notebook APIs. The implementation follows the existing CLI architecture patterns — specifically mirroring the catalog import/export approach — with two new helper modules (`notebook_export.py` and `notebook_import.py`) and integration into the existing `bundle`, `deploy`, `destroy`, and `dry-run` commands.
 
 ### Design Goals
 
 - **Consistency**: Follow the same module structure, error handling, and reporting patterns established by the catalog import/export feature
 - **Resilience**: Individual notebook failures do not block the entire export/import operation; failures are collected and reported
-- **Upsert Safety**: Import uses a create-or-replace strategy — old versions are only deleted after the new version is confirmed ACTIVE and updated
+- **In-Place Updates**: `StartNotebookSync` handles both create and update — no delete-old-version flow. Run history is preserved across deployments
+- **Metadata Tracking**: Target notebooks are tagged with `smus-cicd-source-notebook-id` to maintain source-to-target mapping via `UpdateNotebook`
 - **Idempotency**: Import operations use client tokens to prevent duplicate notebook creation on retries
+- **Fail-Fast Validation**: When `notebook_ids` is specified, all IDs are validated upfront via `GetNotebook` before any exports begin
 - **Extensibility**: The `content.notebooks` manifest section and `deployment_configuration.notebooks` section follow the same convention as `content.catalog` and `deployment_configuration.catalog`
 
 ### Key Design Decisions
 
-1. **Separate helper modules** rather than embedding logic in command files — follows `catalog_export.py` / `catalog_import.py` pattern
-2. **JSON manifest inside the bundle** (`notebooks/notebook_export_manifest.json`) — mirrors `catalog/catalog_export.json` approach
-3. **Async polling with exponential backoff** for both export (GetNotebookExport) and post-import status (GetNotebook) — avoids hammering APIs
-4. **Notebook ID used directly for file paths** — the notebook ID (pattern: `[a-zA-Z0-9_-]{1,36}`) is inherently safe for both filesystem and S3 key usage, eliminating the need for name sanitization
-5. **Add `notebook` resource type** to the existing `DEPLOY_RESOURCE_TYPES` / `DESTROY_SUPPORTED_RESOURCE_TYPES` registries for destroy support
-6. **Upsert strategy (create-or-replace)** — StartNotebookImport always creates a new notebook (duplicate names allowed), then old version with same name is deleted only after new one is ACTIVE + metadata-updated. This ensures zero data loss.
-7. **Name-only filtering** — `include_names` and `exclude_names` match by notebook name only (case-sensitive exact match), no ID matching
+1. **StartNotebookSync replaces StartNotebookImport** — single API for both create (no notebookId) and update (with notebookId). No delete-old-version flow. Preserves run history.
+2. **Notebook ID-only operation** — manifest uses `notebook_ids` list (optional). No include_names/exclude_names. No name-based filtering anywhere.
+3. **Separate helper modules** rather than embedding logic in command files — follows `catalog_export.py` / `catalog_import.py` pattern
+4. **JSON manifest inside the bundle** (`notebooks/notebook_export_manifest.json`) — mirrors `catalog/catalog_export.json` approach
+5. **Async polling with exponential backoff** for export (GetNotebookExport) — avoids hammering APIs
+6. **Notebook ID used directly for file paths** — the notebook ID (pattern: `[a-zA-Z0-9_-]{1,36}`) is inherently safe for both filesystem and S3 key usage
+7. **Add `notebook` resource type** to the existing `DEPLOY_RESOURCE_TYPES` / `DESTROY_SUPPORTED_RESOURCE_TYPES` registries for destroy support
+8. **Metadata-based source tracking** — deploy discovers existing target notebooks via ListNotebooks + GetNotebook, checks for `smus-cicd-source-notebook-id` in metadata
+9. **No unique name enforcement** — IDs are unique by definition; duplicate names are allowed in DataZone
+10. **UpdateNotebook failure = FAILED** — no separate warning category; if metadata/config porting fails, the notebook counts as failed
+
 
 ---
 
@@ -32,35 +38,43 @@ This feature adds native Data Notebooks support to the SMUS CI/CD CLI, enabling 
 graph TD
     subgraph BundlePhase["Bundle Phase (source project)"]
         A[Bundle Command] --> B[NotebookExporter]
-        B --> C[ListNotebooks API]
-        B --> D[GetNotebook API]
-        B --> E[StartNotebookExport API]
-        B --> F[GetNotebookExport API - polling]
+        B --> B1{notebook_ids specified?}
+        B1 -->|Yes| B2[Validate ALL IDs via GetNotebook]
+        B2 -->|Any invalid| B3[FAIL immediately — list invalid IDs]
+        B2 -->|All valid| B4[Export specified notebooks]
+        B1 -->|No| B5[ListNotebooks API — all active]
+        B5 --> B4
+        B4 --> E[StartNotebookExport API per notebook]
+        E --> F[GetNotebookExport — poll with backoff]
         F --> G[Download .ipynb from S3]
         G --> H[Write to notebooks/ in bundle ZIP]
-        D --> I[Write notebook_export_manifest.json]
+        B2 --> I[GetNotebook retrieves parameters/metadata/envConfig]
+        I --> J[Write notebook_export_manifest.json]
     end
 
-    subgraph DeployPhase["Deploy Phase (target project) - Upsert"]
-        J[Deploy Command] --> K[NotebookImporter]
-        K --> K1[ListNotebooks - find old versions by name]
-        K1 --> L[Upload .ipynb to target S3]
-        L --> M[StartNotebookImport API - always creates new]
-        M --> N[Poll GetNotebook until ACTIVE]
-        N --> O[UpdateNotebook API - metadata porting]
-        O --> O1[DeleteNotebook API - remove old version]
-        O1 --> P[Report Summary: new/updated/failed]
+    subgraph DeployPhase["Deploy Phase (target project)"]
+        K[Deploy Command] --> L[NotebookImporter]
+        L --> L1["Step 1: Upload .ipynb files to S3"]
+        L1 --> L2["Step 2: ListNotebooks + GetNotebook each → read metadata"]
+        L2 --> L3["Step 3: Build source→target map from metadata"]
+        L3 --> L4["Step 4: For each manifest entry → StartNotebookSync"]
+        L4 --> L5["Step 5: UpdateNotebook (name, desc, metadata, params, envConfig)"]
+        L5 --> L6["Step 6: Report created/updated/failed"]
     end
 
     subgraph DestroyPhase["Destroy Phase (target project)"]
-        Q[Destroy Command] --> R[ListNotebooks API]
-        R --> R1[Apply include_names/exclude_names filters]
-        R1 --> S[Display destruction plan]
-        S --> T[DeleteNotebook API per notebook]
+        Q[Destroy Command] --> R["ListNotebooks + GetNotebook each"]
+        R --> R1["Filter: has smus-cicd-source-notebook-id metadata"]
+        R1 --> R2{"notebook_ids in manifest?"}
+        R2 -->|Yes| R3["Additionally filter by source IDs in list"]
+        R2 -->|No| R4["Include all with metadata key"]
+        R3 --> S[Display destruction plan]
+        R4 --> S
+        S --> T[DeleteNotebook per notebook]
     end
 
-    H --> |bundle.zip| J
-    I --> |bundle.zip| J
+    H --> |bundle.zip| K
+    J --> |bundle.zip| K
 ```
 
 ### Module Placement
@@ -69,7 +83,7 @@ graph TD
 src/smus_cicd/
 ├── helpers/
 │   ├── notebook_export.py          # NEW: Notebook export logic for bundle
-│   ├── notebook_import.py          # NEW: Notebook import logic for deploy
+│   ├── notebook_import.py          # NEW: Notebook sync logic for deploy
 │   └── ...
 ├── commands/
 │   ├── bundle.py                   # MODIFIED: call notebook_export when enabled
@@ -88,9 +102,9 @@ src/smus_cicd/
 | Command | Integration Point | Action |
 |---------|-------------------|--------|
 | `bundle` | After catalog export | Call `export_notebooks()` if `content.notebooks.enabled` |
-| `deploy` | After catalog import, before bootstrap | Call `import_notebooks()` if manifest present |
-| `destroy` | Validation phase | Call `ListNotebooks` to discover notebooks, apply `include_names`/`exclude_names` filters |
-| `destroy` | Execution phase | Call `DeleteNotebook` for each filtered notebook |
+| `deploy` | After catalog import, before bootstrap | Call `sync_notebooks()` if manifest present and not disabled |
+| `destroy` | Validation phase | ListNotebooks + GetNotebook → filter by metadata → optional `notebook_ids` filter |
+| `destroy` | Execution phase | DeleteNotebook for each filtered notebook |
 | `dry-run` | After catalog checker | `NotebookChecker.check()` validates prerequisites |
 
 ---
@@ -106,8 +120,8 @@ New dataclass for the `content.notebooks` section:
 class NotebookConfig:
     """Notebook export configuration for bundle."""
     enabled: bool = False
-    include_names: Optional[List[str]] = None  # notebook names only (case-sensitive exact match)
-    exclude_names: Optional[List[str]] = None  # notebook names only (case-sensitive exact match)
+    notebook_ids: Optional[List[str]] = None  # Optional list of notebook IDs to export
+    # Pattern per ID: [a-zA-Z0-9_-]{1,36}
 ```
 
 **Parsing**: Added to `ContentConfig`:
@@ -142,57 +156,62 @@ def export_notebooks(
     domain_id: str,
     project_id: str,
     region: str,
-    include_names: Optional[List[str]] = None,
-    exclude_names: Optional[List[str]] = None,
+    notebook_ids: Optional[List[str]] = None,
     polling_timeout: int = 300,
-) -> Tuple[List[ExportedNotebook], NotebookExportManifest]:
+) -> Tuple[List[ExportedNotebook], Dict[str, Any]]:
     """
     Export notebooks from a DataZone project.
+
+    If notebook_ids is specified:
+      1. Validate ALL IDs upfront via GetNotebook (fail-fast)
+      2. If any ID is invalid, fail immediately listing all invalid IDs
+      3. If all valid, export only those notebooks
+
+    If notebook_ids is omitted:
+      1. Call ListNotebooks to discover all active notebooks
+      2. Export all discovered notebooks
 
     Args:
         domain_id: DataZone domain identifier
         project_id: DataZone project identifier
         region: AWS region
-        include_names: Optional list of notebook names to include (case-sensitive exact match)
-        exclude_names: Optional list of notebook names to exclude (case-sensitive exact match)
+        notebook_ids: Optional list of specific notebook IDs to export
         polling_timeout: Max seconds to wait per notebook export (default 300)
 
     Returns:
-        Tuple of (list of exported notebook file contents, manifest object)
+        Tuple of (list of exported notebook objects, manifest dict)
 
     Raises:
         Exception: If ListNotebooks API fails entirely
+        SystemExit: If any notebook_ids are invalid (fail-fast)
     """
 ```
 
 **Internal functions:**
 
 ```python
+def _validate_notebook_ids(
+    client, domain_id: str, notebook_ids: List[str],
+) -> Tuple[List[Dict], List[str]]:
+    """Validate ALL notebook IDs upfront via GetNotebook.
+    Returns (valid_notebooks_with_details, invalid_ids).
+    Calls GetNotebook for EACH ID — retrieves parameters, metadata, envConfig."""
+
 def _list_all_notebooks(client, domain_id: str, project_id: str) -> List[Dict]:
-    """List all active notebooks with pagination."""
-
-def _apply_filters(
-    notebooks: List[Dict],
-    include_names: Optional[List[str]],
-    exclude_names: Optional[List[str]],
-) -> List[Dict]:
-    """Apply include_names/exclude_names filters by notebook name only.
-    Include first, then exclude. Case-sensitive exact match."""
-
-def _matches_name_filter(notebook: Dict, filter_names: List[str]) -> bool:
-    """Check if notebook name matches any entry in the filter list (case-sensitive exact match)."""
+    """List all active notebooks with pagination (follows nextToken)."""
 
 def _export_single_notebook(
     client, s3_client, domain_id: str, project_id: str,
     notebook: Dict, polling_timeout: int,
 ) -> Optional[ExportedNotebook]:
-    """Export a single notebook, poll for completion, download file."""
+    """Export a single notebook: StartNotebookExport → poll → download."""
 
 def _poll_export_status(
     client, domain_id: str, export_id: str, notebook_id: str,
     polling_timeout: int,
 ) -> Optional[str]:
-    """Poll GetNotebookExport with exponential backoff. Returns S3 URI or None."""
+    """Poll GetNotebookExport with exponential backoff + jitter.
+    Initial: 1s, doubles each poll, capped at 30s. Returns S3 URI or None."""
 
 def _build_export_manifest(
     exported: List[ExportedNotebook],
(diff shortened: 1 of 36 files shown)


.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+
+The export process uses `ListNotebooks` to discover notebooks in the source project, `GetNotebook` to retrieve each notebook's full metadata (including parameters, metadata, and environmentConfiguration), `StartNotebookExport` to initiate an asynchronous export of each notebook to IPYNB format, and `GetNotebookExport` to poll for completion and retrieve the S3 output location. The exported `.ipynb` files are included in the bundle archive. During deployment, the import follows an upsert (create-or-replace) strategy: existing notebooks with the same name in the target project are detected via `ListNotebooks`, the new notebook is always created via `StartNotebookImport` (since duplicate names are allowed), polled until ACTIVE via `GetNotebook`, updated with metadata via `UpdateNotebook`, and only then is the old notebook (if any) deleted via `DeleteNotebook`. This ensures zero data loss — the old version is only removed after the new version is confirmed working.
```
vasu2856: Doesn't deleting an existing notebook drop all the related runs? We should not create a new notebook. Instead, figure out a way to update it.

.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+
+**Out of Scope:** Notebook schedules are explicitly out of scope for this iteration. No public API for schedule management currently exists; this will be revisited with the team in a future iteration.
```
vasu2856, line 9: Lets catchup with PM to confirm if this is acceptable. Technically, we have the necessary public APIs from Datazone and Event Bridge Scheduler to achieve this.

.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+- **Notebook_Exporter**: The component responsible for listing project notebooks via `ListNotebooks`, exporting them via `StartNotebookExport`, polling via `GetNotebookExport`, and downloading the exported `.ipynb` files
+- **Notebook_Importer**: The component responsible for uploading `.ipynb` files to S3, performing upsert operations (create new notebook via `StartNotebookImport`, poll until ACTIVE, apply metadata via `UpdateNotebook`, then delete old notebook with same name if one existed)
+- **DataZone_Domain**: An Amazon DataZone domain that contains projects and notebook resources
+- **DataZone_Project**: A project within a DataZone domain that owns notebook resources
+- **SMUS_Notebook**: A notebook resource natively managed by SageMaker Unified Studio, identified by a notebook ID and owned by a project
+- **Notebook_Export**: An asynchronous operation initiated by `StartNotebookExport` that exports a notebook to a specified file format and stores the output in S3
+- **Notebook_Import**: An operation initiated by `StartNotebookImport` that creates a notebook resource in a project from an S3 source location (always creates a new notebook; duplicate names are allowed)
```
vasu2856: Lets discuss about this.

.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+
+**User Story:** As a developer, I want to configure notebook export in my manifest.yaml, so that the bundle command knows which notebooks to export from my SMUS project.
+
+#### Acceptance Criteria
+
+1. THE Manifest SHALL support a `content.notebooks` section to configure notebook export
+2. THE `content.notebooks` section SHALL support an `enabled` boolean field to enable or disable notebook export (default: false)
+3. IF `content.notebooks.enabled` is set to true and no `include_names` filter is specified, THEN THE Bundle_Command SHALL export all active notebooks owned by the source project
+4. THE `content.notebooks` section SHALL support an optional `include_names` field containing a list of notebook name strings to export, where each entry is a notebook name (up to 256 characters, case-sensitive exact match)
+5. IF `content.notebooks.include_names` is specified, THEN THE Bundle_Command SHALL match each entry against the notebook name (case-sensitive exact match) and export notebooks that match
```
vasu2856: Are the notebook names required to be unique? If not, how are we handling collisions? isn't ID better?

.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+6. THE `content.notebooks` section SHALL support an optional `exclude_names` field containing a list of notebook name strings to exclude from export, where each entry is a notebook name (up to 256 characters, case-sensitive exact match)
+7. IF `content.notebooks.exclude_names` is specified, THEN THE Bundle_Command SHALL match each entry against the notebook name (case-sensitive exact match) and exclude notebooks that match
```
vasu2856: This is a bit too complicated. Can we just start with the filter to enable notebook import and add this based on future feature requests?

.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+7. IF the export status is `FAILED`, THEN THE Notebook_Exporter SHALL log the error message from the `error` field, count the notebook as failed, and continue with the next notebook
+8. THE Notebook_Exporter SHALL store each exported `.ipynb` file in a `notebooks/` directory within the bundle archive, using the notebook's source identifier as the filename (e.g., `notebooks/{sourceNotebookId}.ipynb`)
+9. THE Notebook_Exporter SHALL produce a `notebooks/notebook_export_manifest.json` metadata file in the bundle containing the list of exported notebooks with their source IDs, names, descriptions, file paths within the bundle, and the `parameters`, `metadata`, and `environmentConfiguration` fields retrieved from the GetNotebook_API
+10. THE Notebook_Exporter SHALL implement a configurable polling timeout with a default of 300 seconds per notebook to avoid indefinite waiting on stuck export operations
+11. IF the polling timeout is exceeded for a notebook export, THEN THE Notebook_Exporter SHALL log a warning including the notebook name and elapsed time, count the notebook as failed, and continue with the next notebook
```
vasu2856: If notebook export is going to create new revisions, this should be an error and we should fail the deploy. As the next deployment will deploy completely brand new notebooks.

.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Requirement 4: Import Notebooks During Deploy (Upsert)
+
+**User Story:** As a developer, I want the deploy command to import notebooks into the target SMUS project using an upsert strategy, so that exported notebooks appear as native notebook resources in the target environment and existing notebooks with the same name are seamlessly replaced without data loss.
+
+#### Acceptance Criteria
+
+1. WHEN the deploy command processes a bundle containing a `notebooks/notebook_export_manifest.json` file, THE Deploy_Command SHALL invoke the Notebook_Importer after storage deployments
+2. WHEN the Notebook_Importer is invoked, THE Notebook_Importer SHALL upload each exported `.ipynb` file referenced in the Notebook_Export_Manifest from the bundle to the target project's `default.s3_shared` connection S3 URI under a `notebooks/imports/` prefix
+3. WHEN a notebook entry in the Notebook_Export_Manifest references a `filePath` that does not exist in the bundle archive, THE Notebook_Importer SHALL log an error identifying the missing file and notebook name, count it as failed, and continue with the next notebook
```
vasu2856: Why would this happen?

.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+4. BEFORE importing each notebook, THE Notebook_Importer SHALL call the ListNotebooks_API with `owningProjectIdentifier` and `status=ACTIVE` filter on the target project to discover existing notebooks, and identify any notebook with a name matching the notebook being imported as the "old version"
+5. FOR EACH notebook in the Notebook_Export_Manifest whose file was successfully uploaded, THE Notebook_Importer SHALL call the StartNotebookImport_API with the notebook's `name`, `description`, the target project's `owningProjectIdentifier`, the target project's domain identifier as `domainIdentifier`, and the `sourceLocation` pointing to the uploaded S3 file
+6. THE Notebook_Importer SHALL use a client token derived from the notebook name and deployment timestamp, truncated to a maximum of 64 characters, to ensure idempotent import operations (retrying the same deploy produces the same notebook, not a duplicate)
```
vasu2856: What's the purpose of this client token beyond the import?

.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+2. WHEN the Notebook_Importer is invoked, THE Notebook_Importer SHALL upload each exported `.ipynb` file referenced in the Notebook_Export_Manifest from the bundle to the target project's `default.s3_shared` connection S3 URI under a `notebooks/imports/` prefix
+3. WHEN a notebook entry in the Notebook_Export_Manifest references a `filePath` that does not exist in the bundle archive, THE Notebook_Importer SHALL log an error identifying the missing file and notebook name, count it as failed, and continue with the next notebook
+4. BEFORE importing each notebook, THE Notebook_Importer SHALL call the ListNotebooks_API with `owningProjectIdentifier` and `status=ACTIVE` filter on the target project to discover existing notebooks, and identify any notebook with a name matching the notebook being imported as the "old version"
+5. FOR EACH notebook in the Notebook_Export_Manifest whose file was successfully uploaded, THE Notebook_Importer SHALL call the StartNotebookImport_API with the notebook's `name`, `description`, the target project's `owningProjectIdentifier`, the target project's domain identifier as `domainIdentifier`, and the `sourceLocation` pointing to the uploaded S3 file
```
vasu2856: If the notebook already exists, do we need to update these fields and metadata?

.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+
+#### Acceptance Criteria
+
+1. WHEN the deploy command processes a bundle containing a `notebooks/notebook_export_manifest.json` file, THE Deploy_Command SHALL invoke the Notebook_Importer after storage deployments
+2. WHEN the Notebook_Importer is invoked, THE Notebook_Importer SHALL upload each exported `.ipynb` file referenced in the Notebook_Export_Manifest from the bundle to the target project's `default.s3_shared` connection S3 URI under a `notebooks/imports/` prefix
+3. WHEN a notebook entry in the Notebook_Export_Manifest references a `filePath` that does not exist in the bundle archive, THE Notebook_Importer SHALL log an error identifying the missing file and notebook name, count it as failed, and continue with the next notebook
+4. BEFORE importing each notebook, THE Notebook_Importer SHALL call the ListNotebooks_API with `owningProjectIdentifier` and `status=ACTIVE` filter on the target project to discover existing notebooks, and identify any notebook with a name matching the notebook being imported as the "old version"
```
vasu2856: What's missing is what happens once we identify an existing notebook. I would like to know whether we will update all the metadata  or if we will compute the delta between the metadata and update accordingly. In case we update all, want to confirm that there are no side effects. In case, we do a delta, want to know how we are planning to keep track of the diff.

.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+5. FOR EACH notebook in the Notebook_Export_Manifest whose file was successfully uploaded, THE Notebook_Importer SHALL call the StartNotebookImport_API with the notebook's `name`, `description`, the target project's `owningProjectIdentifier`, the target project's domain identifier as `domainIdentifier`, and the `sourceLocation` pointing to the uploaded S3 file
+6. THE Notebook_Importer SHALL use a client token derived from the notebook name and deployment timestamp, truncated to a maximum of 64 characters, to ensure idempotent import operations (retrying the same deploy produces the same notebook, not a duplicate)
+7. WHEN the StartNotebookImport_API returns a notebook identifier, THE Notebook_Importer SHALL poll the GetNotebook_API with exponential backoff (starting at 1 second, doubling on each poll, capped at 10 seconds) until the newly created notebook's status is `ACTIVE`
+8. IF the newly created notebook does not reach `ACTIVE` status within 120 seconds, THEN THE Notebook_Importer SHALL log a warning including the notebook name and elapsed time, count the notebook as failed, and continue with the next notebook
```
vasu2856: is 120 secs typical and sufficient? 2 mins seems smaller for decent sized notebooks. Also, can customers update this timeout value?

.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+9. WHEN the newly created notebook status is `ACTIVE`, THE Notebook_Importer SHALL call the UpdateNotebook_API with the target project's `domainIdentifier`, the newly imported notebook's identifier, `owningProjectIdentifier`, and the `parameters`, `metadata`, and `environmentConfiguration` values from the manifest
+10. IF an old version notebook with the same name was identified in step 4, THEN THE Notebook_Importer SHALL delete the old notebook via the DeleteNotebook_API only after the new notebook has reached ACTIVE status and the UpdateNotebook_API call has completed (successfully or with a warning)
```
vasu2856: If we are doing an upsert, why do we have to delete the existing notebook? Also, what happens to the history - prior runs, schedule, etc

Also, do we also keep track of the library management as part of the metadata in the manifest?

.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+11. IF the DeleteNotebook_API call for the old version fails, THEN THE Notebook_Importer SHALL log a warning including the old notebook name and identifier and error details, but SHALL NOT count the overall notebook import as failed (the new notebook is already active)
```
vasu2856: this should be a failure. Also, will be above delete step clean up such duplicate notebooks as well? Ideally, if we retry import, it should drop both the original notebook and the duplicate we created due to failed deployment.

.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+3. IF the S3 upload fails for a notebook file, THEN THE Notebook_Importer SHALL log the error including the notebook name and S3 destination key, skip the import for that notebook, and continue processing remaining notebooks
+4. IF the target project's `default.s3_shared` connection does not exist or does not contain an `s3Uri` value, THEN THE Notebook_Importer SHALL raise an error indicating the missing connection and skip all notebook imports for that project
+
+### Requirement 7: Error Handling and Resilience
+
+**User Story:** As a developer, I want robust error handling during notebook export and import, so that partial failures do not block the entire deployment.
+
+#### Acceptance Criteria
+
+1. IF the ListNotebooks_API returns an error during export, THEN THE Notebook_Exporter SHALL raise an exception with an error message including the domain identifier, project identifier, and the error response from the API
```
vasu2856: This should fail the deploy command on the whole, as list failure would apply for all notebooks.

.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+2. IF the source project has no notebooks matching the configured filters, THEN THE Notebook_Exporter SHALL produce an empty Notebook_Export_Manifest with zero notebooks and log an informational message indicating the project identifier and the filters that were applied
```
vasu2856: Why do you need an empty manifest? We should either have a manifest with other metadata like search time, filters etc for debugging purposes or skip the manifest alltogether

.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+3. IF a StartNotebookExport_API call fails for a specific notebook during export, THEN THE Notebook_Exporter SHALL log the notebook name and error, then continue with the next notebook
+4. IF the Notebook_Export_Manifest is missing the top-level `metadata` or `notebooks` keys, or if `metadata` is missing any of `sourceProjectId`, `sourceDomainId`, `exportTimestamp`, or `notebookCount`, THEN THE Notebook_Importer SHALL raise a validation error before attempting any API calls
```
vasu2856, line 145: shouldn't this short circuit the notebook import flow skip it?

.kiro/specs/data-notebooks-support/requirements.md
```diff
@@ -0,0 +1,208 @@
...
+5. IF any notebook imports or updates fail, THEN THE Deploy_Command SHALL fail the overall deployment with a non-zero exit code and output a failure message listing the names of all notebooks that failed and their respective error messages
+6. THE Notebook_Exporter SHALL implement exponential backoff with jitter when polling GetNotebookExport_API, starting at an initial interval of 1 second, doubling on each poll, up to a maximum interval of 30 seconds per poll
+7. IF a ThrottlingException is received from any DataZone API call, THEN THE component SHALL retry the request with exponential backoff starting at 1 second and doubling on each retry, up to a maximum of 3 retries
+
+### Requirement 8: Dry Run Validation for Notebooks
```
vasu2856, line 151: is there a way to verify if the import will be successful/ fail without actually doing it?

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: Using StartNotebookSync API for this

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: Okay, will talk to Ramesh about this in the cicd channel.

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: This will have StartNotebookSync on the update now

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: Yeah, I think we will switch to IDs

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: There is actually not too much code to support allow and deny listing

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: Yes, it will fail the overall deploy, but I think after it deploys the rest of the notebooks. I can make it fail fast here as well, if that's what we want

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: I guess when the bundle is modified externally

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: I think it is just for the import

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: N/A for new API

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: So, in the subsequent UpdateNotebook call, we will replace all metadata, parameters, description, name, etc. with whatever was defined in the source project. No delta is kept.

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: No, it's arbitrary and customers cannot update it. We can set to to 5 mins, if you feel like it is not sufficient

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: Where do we store library management for the notebook? If it is inside metadata or parameters, it should be copied over

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: N/A for new design

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: I think it does, I will check

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: Fair, we can skip it

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: Well, if the bundle has some notebooks which are invalid, should not we throw and error?

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: Tracking this as a separate AI waiting for scheduling DZ API

.kiro/specs/data-notebooks-support/requirements.md
vasu2856: The question was about #3. shouldn't it fail right away, if we can't initiate an export?

.kiro/specs/data-notebooks-support/design.md
```diff
@@ -2,25 +2,31 @@
...
+This feature adds native Data Notebooks support to the SMUS CI/CD CLI, enabling promotion of SageMaker Unified Studio notebooks across environments (dev → test → prod) using the DataZone Notebook APIs. The implementation follows the existing CLI architecture patterns — specifically mirroring the catalog import/export approach — with two new helper modules (`notebook_export.py` and `notebook_import.py`) and integration into the existing `bundle`, `deploy`, `destroy`, and `dry-run` commands.
 
 ### Design Goals
 
 - **Consistency**: Follow the same module structure, error handling, and reporting patterns established by the catalog import/export feature
 - **Resilience**: Individual notebook failures do not block the entire export/import operation; failures are collected and reported
-- **Upsert Safety**: Import uses a create-or-replace strategy — old versions are only deleted after the new version is confirmed ACTIVE and updated
+- **In-Place Updates**: `StartNotebookSync` handles both create and update — no delete-old-version flow. Run history is preserved across deployments
```
vasu2856, line 11: we should not talk about the 'delete old version' stuff any more.

.kiro/specs/data-notebooks-support/requirements.md
Shnekit: I think currently, it will go ahead and deploy what it can and then fail the deployment stating the notebooks that were skipped and why. Do you think we should fail fast instead?

.kiro/specs/data-notebooks-support/design.md
```diff
@@ -2,25 +2,31 @@
...
+- **Metadata Tracking**: Target notebooks are tagged with `smus-cicd-source-notebook-id` to maintain source-to-target mapping via `UpdateNotebook`
```
vasu2856, line 12: is the source notebook id from lower stage? But how do we know that the source notebook id is the same? I would not want to carry any info from the lower stage into the resource. We should keep it at the bundle layer.

.kiro/specs/data-notebooks-support/design.md
```diff
@@ -2,25 +2,31 @@
...
-2. **JSON manifest inside the bundle** (`notebooks/notebook_export_manifest.json`) — mirrors `catalog/catalog_export.json` approach
-3. **Async polling with exponential backoff** for both export (GetNotebookExport) and post-import status (GetNotebook) — avoids hammering APIs
-4. **Notebook ID used directly for file paths** — the notebook ID (pattern: `[a-zA-Z0-9_-]{1,36}`) is inherently safe for both filesystem and S3 key usage, eliminating the need for name sanitization
-5. **Add `notebook` resource type** to the existing `DEPLOY_RESOURCE_TYPES` / `DESTROY_SUPPORTED_RESOURCE_TYPES` registries for destroy support
-6. **Upsert strategy (create-or-replace)** — StartNotebookImport always creates a new notebook (duplicate names allowed), then old version with same name is deleted only after new one is ACTIVE + metadata-updated. This ensures zero data loss.
-7. **Name-only filtering** — `include_names` and `exclude_names` match by notebook name only (case-sensitive exact match), no ID matching
+1. **StartNotebookSync replaces StartNotebookImport** — single API for both create (no notebookId) and update (with notebookId). No delete-old-version flow. Preserves run history.
```
vasu2856, line 19: Do we have to talk about StartNotebookImport anymore?

.kiro/specs/data-notebooks-support/design.md
```diff
@@ -2,25 +2,31 @@
...
+2. **Notebook ID-only operation** — manifest uses `notebook_ids` list (optional). No include_names/exclude_names. No name-based filtering anywhere.
+3. **Separate helper modules** rather than embedding logic in command files — follows `catalog_export.py` / `catalog_import.py` pattern
+4. **JSON manifest inside the bundle** (`notebooks/notebook_export_manifest.json`) — mirrors `catalog/catalog_export.json` approach
+5. **Async polling with exponential backoff** for export (GetNotebookExport) — avoids hammering APIs
+6. **Notebook ID used directly for file paths** — the notebook ID (pattern: `[a-zA-Z0-9_-]{1,36}`) is inherently safe for both filesystem and S3 key usage
+7. **Add `notebook` resource type** to the existing `DEPLOY_RESOURCE_TYPES` / `DESTROY_SUPPORTED_RESOURCE_TYPES` registries for destroy support
```
vasu2856, line 25: I'm assuming dry run works too?

.kiro/specs/data-notebooks-support/design.md
```diff
@@ -2,25 +2,31 @@
...
+8. **Metadata-based source tracking** — deploy discovers existing target notebooks via ListNotebooks + GetNotebook, checks for `smus-cicd-source-notebook-id` in metadata
```
vasu2856, line 26: yeah. We should discuss this.

.kiro/specs/data-notebooks-support/design.md
Shnekit: Yes, I think it was mentioned somewhere as well, apologies if it is not

.kiro/specs/data-notebooks-support/design.md
Shnekit: I think we do not need it anymore. StartNotebookSync works for both creating and updating the notebooks depending on whether you give it a notebook id

.kiro/specs/data-notebooks-support/design.md
Shnekit: The assumption here is that we are tagging all the resources with the source notebook id taken from the bundle. So, if the bundle was taken from dev, then test and prod notebooks will be tagged with dev notebook id as a source. The source notebook id will stay the same unless it is recreated, in which case we are recreating a new notebook in test and prod as well.
I do not really see any way of keeping this information in the bundle since it is only dealing with source creds. So, there is no way for us to figure out which notebook in the target corresponds to the which in the source because the notebooks ids will always be different across stages.
We could map everything by the name, instead of the id, but we will have to force the user to keep unique names and everything breaks, if they accidentally change any name, so I think it will be more error-prone.
Let's discuss this further.

.kiro/specs/data-notebooks-support/design.md
Shnekit: Sure, I can clean it up

src/smus_cicd/commands/deploy.py
```diff
@@ -1357,6 +1370,169 @@ def _validate_deployed_workflows(
...
+
+    try:
+        with zipfile.ZipFile(local_bundle_path, "r") as zip_ref:
+            namelist = zip_ref.namelist()
+
+            if "notebooks/notebook_export_manifest.json" not in namelist:
+                # No notebook manifest — skip silently (backward compatible)
+                return True
+
+            # ── Extract manifest ──────────────────────────────────────────
```
vasu2856, line 1440: Why do we have to unzip and extract manifest in the notebook path? don't we already do it once as part of deploy and every individual resource import works off it?

src/smus_cicd/commands/deploy.py
```diff
@@ -1357,6 +1370,169 @@ def _validate_deployed_workflows(
...
+        if not project_id:
+            typer.echo(
+                f"❌ Could not find project ID for project '{project_name}' "
+                "— skipping notebook sync",
+                err=True,
+            )
+            return False
+
+        # ── Resolve S3 URI from default.s3_shared connection ─────────────
+        from ..helpers.connections import get_project_connections
```
vasu2856: Why is all this repeated for notebook? we should just do this once for a target from the manifest as a common utility and reuse it through out

src/smus_cicd/helpers/destroy_executor.py
```diff
@@ -489,6 +489,88 @@ def _log(msg: str) -> None:
                     )
                 )
 
+    # -----------------------------------------------------------------------
+    # Step f.5: Delete notebooks (CI/CD-managed only, identified by metadata)
```
vasu2856: How do we enforce CI/CD managed notebooks only? I don't see any tag based filter

src/smus_cicd/helpers/destroy_validator.py
```diff
@@ -36,6 +36,104 @@
...
+    notebook_ids_filter=None,
+    _dz_client_override=None,
+):
+    """
+    Discover target notebooks that were deployed by this CI/CD tool.
+
+    Steps:
+      1. ListNotebooks with owningProjectIdentifier + status=ACTIVE (paginated).
+      2. GetNotebook for each to read metadata.
+      3. Include only notebooks whose metadata contains ``smus-cicd-source-notebook-id``.
```
vasu2856, line 52: If customers ever run this method on dev stage,we won't have 'smus-cicd-source-notebook-id' tag and hence we don't touch the ones we didn't create?

src/smus_cicd/commands/deploy.py
Shnekit: I think this is talking about notebook-export-manifest.yaml. That is a separate metadata file we create for notebooks which has not been extracted anywhere else prior to this

src/smus_cicd/commands/deploy.py
Shnekit: Will refactor the code to make sure we are not redoing steps

src/smus_cicd/helpers/destroy_validator.py
Shnekit: Yes. Customers are not supposed to run destroy on dev anyway because destroy is supposed to delete stuff create by SMUS CI/CD. Since there is nothing SMUS CI/CD created, tried to avoid deleting anything there either

src/smus_cicd/commands/deploy.py
vasu2856: But why do we extract separately? shouldn't we extract once and use it for the entire deploy flow for all resources?

Shnekit: I think that is how the deploy command is structured in general. `_deploy_files_from_bundle ` and `_import_catalog_from_bundle ` are also extracting their own stuff separately. 

I can add an AI to make it extract everything at the top instead. But I do not think it should be a blocker for this PR

vasu2856: yes. Please add a backlog item to extract once for a deploy and all the sub modules use it.

vasu2856 (approved):
LGTM. Ship it :)

src/smus_cicd/commands/deploy.py
Shnekit: https://app.asana.com/1/8442528107068/project/1211853550096664/task/1217909171453976?focus=true

PR #266, opened 2026-07-02
fix(workflows): Mitigate script injection in GitHub Actions

Shnekit:
## Summary

Moves user-controlled event context data and job outputs from inline `${{ }}` expressions in `run:` blocks to intermediate environment variables. This prevents potential shell command injection by ensuring untrusted values are treated as data rather than part of the generated shell script.

## Changes

| Workflow | Data moved to env |
|----------|-------------------|
| `slack-pr-notification.yml` | `github.event.pull_request.html_url`, `user.login` |
| `mwaa-pipeline-lifecycle.yml` | `github.event.inputs.domain_name`, `project_name`, `needs.setup.outputs.*` |
| `serverless-pipeline-lifecycle.yml` | Same as mwaa |
| `e2e-mlops-promote.yml` | `github.event.client_payload.model_package_arn`, `inputs.*`, `needs.prepare.outputs.*` |
| `e2e-mlops-deploy.yml` | `inputs.stage`, `inputs.model_package_arn`, `needs.resolve.outputs.*` |

## Context

Per GitHub's security guidance on [script injections](https://securitylab.github.com/resources/github-actions-untrusted-input/), expressions like `${{ github.event.inputs.* }}` in `run:` blocks are substituted into the shell script _before_ execution. If the value contains shell metacharacters, it could break out of the string context and execute arbitrary commands. Routing through `env:` makes the shell treat the value as data.

## Testing

- All YAML files validated with `yaml.safe_load()`
- No functional behavior change — same values, just passed through env vars

## Related

- AppSec ticket: https://t.corp.amazon.com/V2270725343

### DIFF
--- a/.github/workflows/e2e-mlops-deploy.yml
+++ b/.github/workflows/e2e-mlops-deploy.yml
@@ -58,8 +58,11 @@ jobs:
     steps:
       - id: set
         shell: bash
+        env:
+          INPUT_STAGE: ${{ inputs.stage }}
+          INPUT_MODEL_PACKAGE_ARN: ${{ inputs.model_package_arn }}
         run: |
-          STAGE="${{ inputs.stage }}"
+          STAGE="$INPUT_STAGE"
           STAGE="${STAGE:-dev}"
 
           if [ "$STAGE" = "all" ]; then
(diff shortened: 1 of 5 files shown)


vasu2856 (approved):
LGTM

PR #273, opened 2026-07-09
fix(security): Prevent env var secret exfiltration via S3 workflow YAMLs

Shnekit:
## Summary

Prevents secrets defined via `environment_variables` in the manifest from being stored in S3, where data engineers with project bucket access could read them.

## Problem

When the CLI deployed workflows, the `ContextResolver` resolved `{env.VAR}` placeholders to their actual values and uploaded the resolved YAML to S3. This meant secrets (DB passwords, API keys, tokens) stored in CI environment variables were permanently written to the project's S3 bucket in plain text.

## Solution

Instead of resolving `{env.VAR}` to the actual value at deploy time, the resolver now emits Jinja template syntax (`{{ params.VAR }}`). The actual values are passed at **runtime** via MWAA Serverless `OverrideParameters` in the `start_workflow_run` API call.

### Before
```
{env.SECRET_KEY} → "actual-secret-value" (written to S3)
```

### After
```
{env.SECRET_KEY} → "{{ params.SECRET_KEY }}" (written to S3, no secret)
                  → actual value injected at runtime via OverrideParameters
```

## Changes

| File | Change |
|------|--------|
| `context_resolver.py` | `{env.VAR}` emits `{{ params.VAR }}` instead of resolving |
| `airflow_serverless.py` | `start_workflow_run` accepts `override_parameters` |
| `operations.py` | Passes `environment_variables` as override params at runtime |
| `commands/run.py` | Same for CLI run path |
| Unit tests | Updated assertions to expect Jinja syntax |

## Backward Compatibility

- The `{env.VAR}` syntax in workflow YAMLs remains unchanged
- No manifest or example changes required
- `{proj.*}`, `{domain.*}`, `{stage.*}` still resolve at deploy time (infrastructure metadata, not secrets)

## Testing

- Manually verified with MWAA Serverless: `create-workflow` accepts Jinja placeholders, `start-workflow-run` with `--override-parameters` substitutes them correctly in both operator fields and `input_params`
- All unit tests pass (22/22 context resolver + all substitution tests)
- E2E verified with data-notebooks example: S3 contains only `{{ params.TEST_SECRET }}`, notebook output shows actual value

## Related

- AppSec ticket: https://t.corp.amazon.com/P460824733/communication

### DIFF
--- a/src/smus_cicd/commands/run.py
+++ b/src/smus_cicd/commands/run.py
@@ -585,7 +585,18 @@ def _execute_airflow_serverless_workflows(
                 typer.echo(f"🔗 ARN: {workflow_arn}")
 
             # Start workflow run (workflow uses role specified during creation)
-            result = airflow_serverless.start_workflow_run(workflow_arn, region=region)
+            # Pass environment_variables as override parameters for runtime resolution
+            override_parameters = None
+            if (
+                hasattr(target_config, "environment_variables")
+                and target_config.environment_variables
+            ):
+                override_parameters = {
+                    k: str(v) for k, v in target_config.environment_variables.items()
+                }
+            result = airflow_serverless.start_workflow_run(
+                workflow_arn, region=region, override_parameters=override_parameters
+            )
 
             if result.get("success"):
                 run_id = result.get("run_id")
--- a/src/smus_cicd/helpers/context_resolver.py
+++ b/src/smus_cicd/helpers/context_resolver.py
@@ -157,7 +157,7 @@ class ContextResolver:
             values. This prevents secrets from being stored in S3.
 
         Raises:
-            ValueError: If any non-env variable cannot be resolved
+            ValueError: If any variable cannot be resolved
         """
         context = self._build_context()
 
@@ -172,6 +172,9 @@ class ContextResolver:
             # Convert {env.VAR} to Jinja template syntax for runtime resolution
             # instead of resolving to actual values (prevents secret exposure in S3)
             if namespace == "env":
+                if path not in self.env_vars:
+                    unresolved.append(f"{{{namespace}.{path}}}")
+                    return match.group(0)
                 return "{{ params." + path + " }}"
 
             try:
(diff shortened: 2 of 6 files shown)


src/smus_cicd/helpers/context_resolver.py
```diff
@@ -151,10 +151,13 @@ def resolve(self, content: str) -> str:
...
         Returns:
-            Content with variables replaced
+            Content with variables replaced. Note: {env.VAR} references are converted
+            to Jinja template syntax ({{ params.VAR }}) for runtime resolution via
+            MWAA Serverless override parameters, rather than being resolved to actual
+            values. This prevents secrets from being stored in S3.
 
         Raises:
-            ValueError: If any variable cannot be resolved
+            ValueError: If any non-env variable cannot be resolved
```
vasu2856: shouldn't we check for env as well? won't it end up with runtime error, otherwise?

src/smus_cicd/commands/run.py
```diff
@@ -585,7 +585,18 @@ def _execute_airflow_serverless_workflows(
                 typer.echo(f"🔗 ARN: {workflow_arn}")
 
             # Start workflow run (workflow uses role specified during creation)
-            result = airflow_serverless.start_workflow_run(workflow_arn, region=region)
+            # Pass environment_variables as override parameters for runtime resolution
```
vasu2856, line 588: curious why we have these duplicated between operations.py in worklfows folder and here.

Shnekit: This is for `smus-cli run ...` command which let's you directly run the defined workflows.

The workflows/operations.py is where `smus-cli deploy ...` goes when it sees bootstrap action that does workflow.run.

Shnekit: These 2 are currently separated because the logic is a bit different, we could technically merge them together, but that is out of scope of this PR.

src/smus_cicd/helpers/context_resolver.py
Shnekit: Fair. Fixed it

src/smus_cicd/commands/run.py
vasu2856: yeah. Can you keep track of this please.

vasu2856 (approved):
LGTM

src/smus_cicd/commands/run.py
Shnekit: Added a task to Asana

PR #284, opened 2026-07-15
fix(ml): Pin sagemaker SDK v2 for sklearn notebook compatibility

Shnekit:
## Summary

The execution environment (`IMAGE_VERSION: 4.2.2-cpu`) now ships SageMaker Python SDK v3, which completely removed the `sagemaker.sklearn` module and reorganized the SDK (no more `sagemaker.model.Model`, `sagemaker.estimator.Estimator`, etc.). There is no v3 equivalent for sklearn script-mode training/deployment workflows.

## Fix

Add a pip install cell to both ML notebooks that pins `sagemaker>=2.200,<3` before any sagemaker imports. This restores the original `SKLearnModel` and `SKLearn` estimator usage which works correctly with sagemaker v2.

**Changes per notebook:**
- Replace the "Package upgrade skipped" cell with `pip install -q 'sagemaker>=2.200,<3'`
- Restore original `from sagemaker.sklearn import SKLearnModel` (deployment)
- Restore original `from sagemaker.sklearn.estimator import SKLearn` (training)

## Why pin v2 instead of migrating to v3?

SDK v3 uses `ModelTrainer` and `ModelBuilder` which are designed for LLM/foundation model workflows. There is no v3 equivalent for running custom sklearn scripts in framework containers with `entry_point`/`source_dir` patterns. The [official AWS docs](https://docs.aws.amazon.com/sagemaker/latest/dg/sklearn.html) still reference v2-style APIs for sklearn.

## Testing

- All 999 unit tests pass
- Both notebooks validated as correct JSON
- No remaining references to generic `sagemaker.model`/`sagemaker.estimator` imports


Shnekit:
Trying to fix broken tests.

Is not introducing regression.

PR #302, opened 2026-08-06
[workflow_create_handler] Tag MWAA workflows with DataZone project and domain IDs

### LINKED ISSUE #297: MWAA Serverless workflows created by workflow.create are not associated with the DataZone project (missing tags + NetworkConfiguration)
Summary
Workflows created via the workflow.create bootstrap action are created at the account level but are not associated with the target DataZone/SMUS project. They do not appear in the project's Workflows UI. AWS Support confirmed the workflows are not tagged to the project.

Environment
CLI: aws-smus-cicd-cli
project: in Domain account(main Sagemaker domain with associated accounts)
AWS: associated account (tooling and s3)
API: airflow-serverless (MWAA Serverless)
Root Cause (from source inspection)
1. Missing DataZone association tags
In smus_cicd/bootstrap/handlers/workflow_create_handler.py, create_workflow is called with only these tags:
tags={
    "Pipeline": manifest.application_name,
    "Target": target_config.project.name,
    "STAGE": stage_name.upper(),
    "CreatedBy": "SMUS-CICD",
}
None of the SMUS/DataZone association tags are set:

AmazonDataZoneProject = <project_id>
AmazonDataZoneDomain = <domain_id>
(and/or AmazonDataZoneEnvironment)
Without these, SMUS does not scope the workflow to the project, so it's invisible in the project UI.

Shnekit:
## Summary

Fixes #297

MWAA Serverless workflows created by `workflow.create` were not associated with the DataZone project/domain in the console because the required tags were missing.

## Changes

**Add `AmazonDataZoneProject` and `AmazonDataZoneDomain` tags** at workflow creation time so DataZone can display the workflow under the correct project.

**Reconcile tags on pre-existing workflows** via a new `_ensure_workflow_tags()` helper. After every create/update, the handler fetches the workflow's current tags and calls `tag_resource` only for those that are missing. This recovers workflows created before this fix without touching any tags the user may have set manually.

**Fix `STAGE` tag always showing `UNKNOWN`** — the value was previously read from `config` (where `stage_name` is never stored). It now reads from the execution context using the following priority:
1. `target_config.stage` — explicit `stage:` field in the manifest (e.g. `TEST`)
2. `context["stage_name"]` / `context["stage"]` — the target key name (e.g. `test-idc`)

**Single source of truth for tags** — the `workflow_tags` dict is defined once per workflow iteration and passed to both `create_workflow` and `_ensure_workflow_tags`, eliminating the duplicated dict that existed before.

## Testing

Verified with `examples/analytic-workflow/data-notebooks/manifest-idc.yaml` — workflows now appear in IdC domains with all required tags, and `STAGE` correctly shows `TEST` instead of `UNKNOWN`.

### DIFF
--- a/src/smus_cicd/bootstrap/handlers/workflow_create_handler.py
+++ b/src/smus_cicd/bootstrap/handlers/workflow_create_handler.py
@@ -58,7 +58,13 @@ def handle_workflow_create(
     # Get required info from context
     project_name = target_config.project.name
     region = config["region"]
-    stage_name = config.get("stage_name", "unknown")
+    # Prefer the explicit "stage" field from the manifest (e.g. "TEST"), fall back to
+    # the target key name (e.g. "test-idc") which is stored in context as "stage_name"/"stage"
+    stage_name = (
+        getattr(target_config, "stage", None)
+        or context.get("stage_name")
+        or context.get("stage", "unknown")
+    )
 
     # Get project info from metadata (resolved once in deploy)
     project_info = metadata.get("project_info", {})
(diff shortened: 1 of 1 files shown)


vasu2856 (approved):
LGTM

PR #319, opened 2026-08-25
feat(workflow): Inherit Tooling VPC/CMK and IdC log group

### LINKED ISSUE #318: MWAA Serverless workflows missing SMUS Network config and cloudwatch path issue
We are trying to deploy mwaa serverless workflows using the SMUS CLI. And found following configurations misconfigured/missing within workflow:

- The workflow's network configuration (VPC, Subnet, Security group) is missing. When we create workflows manually, network configurations map correctly.
- We have CMK setup in SMUS project, however workflows are created with default AWS KMS keys
- The cloudwatch log group path is misconfigured. We expect path to be of format `/aws/mwaa-serverless/<domain_id>-<project_id>/<workflow_name>`. However the path getting generated is different.
- We would like to add custom tags to workflows being created. I didn't find any way to add them (either through manifest or CLI flag)

Shnekit:
## Summary

CI/CD-created MWAA Serverless workflows now inherit the target project's Tooling blueprint **network** (VPC subnets + security group) and **encryption** (CMK) configuration, and use IdC-namespaced CloudWatch **log groups** — matching what the SageMaker Unified Studio UI does. When the blueprint uses defaults, each setting is omitted so the service default applies.

Closes #318

## What changed

- **Tooling resolution** (`datazone.py`): resolve the default Tooling environment via managed `Tooling.GA` / `Tooling.Lightning` blueprints, with an IAM-connection fallback for projects whose tooling slot is a custom blueprint (mirrors the UI's `CachedProjectEnvironmentsService`).
- **VPC**: derived from a matching domain-scoped VPC connection ("global VPC") when present, else from the tooling environment's provisioned resources (`privateSubnets` / `securityGroup`). **CMK** from `kmsKeyArn`. Each omitted when the blueprint uses defaults.
- **Domain type**: detect IdC vs IAM (`get_domain().singleSignOn`). For IdC domains, set an explicit log group `/aws/mwaa-serverless/<domain-id>-<project-id>/<workflow-name>`. IAM domains keep service-default naming.
- **`create_workflow`** (`airflow_serverless.py`): pass `NetworkConfiguration`, `EncryptionConfiguration`, and `LoggingConfiguration` on create. On **update** (existing workflow), apply network + logging (both mutable) but **never** encryption — `UpdateWorkflow` has no `EncryptionConfiguration` field, so the CMK is fixed at creation. The CMK log line is worded so it does not imply existing workflows are re-encrypted.

## Behavior for existing workflows

- VPC and log group converge on re-deploy (mutable via `UpdateWorkflow`; each update creates a new workflow version).
- CMK cannot be changed on an existing workflow; it must be recreated to adopt a different key.

## Testing

- New `tests/unit/helpers/test_tooling_workflow_config.py`: resolver (VPC-connection vs provisioned-resources branches, defaults), managed-blueprint + IAM-connection fallback resolution, domain-type detection, and create/update parameter wiring (asserts encryption omitted on update).
- Updated handler tests for the new config + IdC log group.
- Full unit suite passes (1016); black + flake8 clean.
- Verified live against a CMK/VPC IdC project: resolver returns the correct subnets/security group/CMK, and the IdC log-group path resolves as expected.

### DIFF
--- a/docs/bootstrap-actions.md
+++ b/docs/bootstrap-actions.md
@@ -60,6 +60,7 @@ stages:
 
 **Properties:**
 - `workflowName` (optional): Specific workflow name to create. Omit to create all workflows defined in `workflows:` section.
+- `tags` (optional): A map of custom `key: value` tags applied to every workflow this action creates, in addition to the tags SMUS CI/CD manages automatically. Keys must be strings. See the note on reserved keys below.
 
 **Use Case:** Create workflows after all required connections exist, allowing workflow definitions to reference connections like MLflow tracking servers.
 
@@ -69,6 +70,29 @@ bootstrap:
   actions:
     - type: workflow.create  # Creates all workflows from workflows: section
 ```
+
+**Example - Custom tags:**
+```yaml
+bootstrap:
+  actions:
+    - type: workflow.create
+      workflowName: ml_training_workflow
+      tags:
+        CostCenter: "1234"
+        Team: analytics
+        Environment: test
+```
+
+Because `tags` lives on the action, tags can differ per stage (each stage has its
+own `workflow.create` action) and per workflow (use separate actions with distinct
+`workflowName` values). Newly created workflows receive the tags immediately;
+existing workflows have any missing tags added on the next deploy.
+
+**Reserved tag keys:** SMUS CI/CD manages these keys on every workflow and they
+cannot be overridden by custom `tags` - supplying any of them fails the deploy
+with a clear error, so rename or remove them:
+`Pipeline`, `Target`, `STAGE`, `CreatedBy`, `AmazonDataZoneDomain`,
+`AmazonDataZoneProject`.
 
 ---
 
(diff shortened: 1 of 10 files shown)


docs/bootstrap-actions.md
```diff
@@ -70,6 +71,29 @@ bootstrap:
...
 ```
 
+**Example - Custom tags:**
+```yaml
+bootstrap:
+  actions:
+    - type: workflow.create
+      workflowName: ml_training_workflow
+      tags:
+        CostCenter: "1234"
```
vasu2856, line 81: Do we also support carry forward the tag and values by keys?

src/smus_cicd/bootstrap/handlers/workflow_create_handler.py
```diff
@@ -37,6 +56,26 @@ def handle_workflow_create(
...
+    custom_tags = action.parameters.get("tags") or {}
+    if custom_tags:
+        if not isinstance(custom_tags, dict):
+            typer.echo(
+                "❌ workflow.create 'tags' must be a mapping of string keys to "
+                "string values"
+            )
+            return False
+        reserved_collisions = sorted(set(custom_tags) & RESERVED_WORKFLOW_TAG_KEYS)
+        if reserved_collisions:
```
vasu2856, line 70: Would this also happen via dry run?

Shnekit: Hmmm, I do not think I included it there. Let me add that

vasu2856 (approved):
LGTM

PR #325, opened 2026-08-26
Add generate-bundle-manifest agent skill

Shnekit:
## Summary

Adds a portable AgentSkills skill, `generate-bundle-manifest`, that helps users generate an SMUS CI/CD deployment manifest (`manifest.yaml`) and, when needed, an MWAA orchestration workflow to replicate a SageMaker Unified Studio project across dev/test/prod stages.

The skill can be imported into an agent environment (e.g. Q CLI, IDE assistants) to scaffold deployment manifests before running the `aws-smus-cicd-cli` bundle/deploy lifecycle.

## Behavior

- **Read-only.** Discovers project resources (connections, S3 storage layout, workflows) and presents the generated manifest in chat. It does not write or modify anything — the user saves and edits the output.
- **Discovers and maps** `.py` scripts, workflow DAGs, notebooks, data, and models to the appropriate `content` sections.
- **Generates dev/test/prod stages** with the current project as the dev stage (real values) and `${VAR}` placeholders for test/prod.
- **Generates a single orchestration workflow** for Glue scripts that are not already referenced by an existing workflow.
- **Catalog is opt-in** (its deploy-time import requires `datazone:AddPolicyGrant`); **notebooks** are included only when `.ipynb` files are discovered.

## Design

Rather than duplicating documentation, the skill references the existing repo docs as the source of truth: `manifest.md` / `manifest-schema.md`, `substitutions-and-variables.md`, `bootstrap-actions.md`, `connections.md`, `cli-commands.md`, `airflow-aws-operators.md`, and `github-actions-integration.md`. For orchestration DAG authoring it also points to the public MWAA Serverless user guide. Orchestration-workflow correctness is best-effort; the primary deliverable is a clean, valid manifest.

## Structure

```
skills/generate-bundle-manifest/
├── SKILL.md
└── references/example-manifests.md
```

## Testing

Manifest generation was validated against a live SMUS project (discovery, manifest generation, bundle, and deploy to a test stage). Orchestration-workflow generation is best-effort as noted above.

### DIFF
--- a/skills/generate-bundle-manifest/SKILL.md
+++ b/skills/generate-bundle-manifest/SKILL.md
@@ -43,11 +43,38 @@ This skill is READ-ONLY: it discovers project resources and presents generated c
 
 ### Step 1: Discover Project Resources (read-only)
 
-Skip if the user already specified exactly what to include. Use your agent's AWS access (e.g. the AWS CLI or an equivalent capability) to discover, scoping every call to the project's domain and project. Obtain the `domainId`/`projectId` from the user, or resolve them by listing (`aws datazone list-domains`, `aws datazone list-projects`).
+Skip if the user already specified exactly what to include. Use your agent's AWS access (the AWS CLI or an equivalent capability) for all discovery. Run the steps in order — step 2 needs the S3 URI resolved in step 1. Do not invent resource or connection names; use only what these calls return.
 
-1. **Connections** — `aws datazone list-connections --domain-identifier <id> --project-identifier <id>`. Reveals S3, Glue, Athena, MWAA, MLflow, Redshift, EMR. See the connections doc linked in References.
-2. **Storage structure** — `aws s3 ls` (or `list-objects-v2` with delimiter `/`) on the shared S3 connection to find prefixes/file types: `src/` (`.py`), `workflows/` (`.yaml`), `data/`, `models/`, `notebooks/` (`.ipynb`).
-3. **MWAA Serverless workflows** — list workflows and post-filter by project tag `AmazonDataZoneProject`.
+First, resolve the domain and project IDs. Ask the user, or list them:
+
+```bash
+aws datazone list-domains
+aws datazone list-projects --domain-identifier <domainId>
+```
+
+**1. Connections (do this first — it yields the S3 location for step 2).**
+
+```bash
+aws datazone list-connections --domain-identifier <domainId> --project-identifier <projectId>
+```
+
+The response lists the project's connections and their types (S3, Glue/SPARK, ATHENA, WORKFLOWS_MWAA, WORKFLOWS_SERVERLESS, MLFLOW, REDSHIFT, EMR). Record each connection NAME (e.g. `default.s3_shared`, `default.workflow_serverless`) — these are the values that go into the manifest's `connectionName` fields. For the shared S3 connection, read its S3 URI from `props.s3Properties.s3Uri` (the bucket is that URI with `s3://` stripped and everything after the first `/` removed). You need this URI for step 2.
+
+**2. Storage structure (uses the S3 URI from step 1).** List the shared S3 location to find top-level prefixes and the file types in each:
+
+```bash
+aws s3 ls <s3Uri-from-step-1>/ --recursive
+```
+
+Classify what you find: `.py` under `src/`, `src/glue-jobs/`, or `jobs/` → Glue/code; `.yaml` under `workflows/` → workflow DAGs; `.ipynb` under `notebooks/` → notebooks; `data/`, `models/` → data/model artifacts.
+
+**3. MWAA Serverless workflows** (only if the project has a `WORKFLOWS_SERVERLESS` connection):
+
+```bash
+aws mwaa-serverless list-workflows
+```
+
+`list-workflows` has no tag filter, so post-filter the results by the project tag `AmazonDataZoneProject` client-side.
 
 Not discoverable this way (manifest-only): git repos, environment variables, IAM role policies, EventBridge events, workflow schedules (a DAG YAML field), QuickSight dashboards.
 
--- a/skills/generate-bundle-manifest/SKILL.md
+++ b/skills/generate-bundle-manifest/SKILL.md
@@ -21,10 +21,10 @@ This skill is READ-ONLY: it discovers project resources and presents generated c
 
 - READ-ONLY: You MUST NOT write, upload, create, or modify any files anywhere. You only read (discovery) and present generated content in the chat; the user saves and edits it.
 - Auto-discover and generate defaults ONLY when the user gives no configuration. If the user specifies resources, stages, or domain/project details, use exactly what they provide and do not add auto-discovered extras.
-- You MUST use `{proj.connection.<name>.<property>}` substitution in generated workflow YAMLs. You MUST NOT hardcode S3 bucket names, IAM role ARNs, or regions. See docs/substitutions-and-variables.md.
+- You MUST use `{proj.connection.<name>.<property>}` substitution in generated workflow YAMLs. You MUST NOT hardcode S3 bucket names, IAM role ARNs, or regions. See the substitutions doc (link in References).
 - All projects referenced in the manifest MUST be pre-created by the user in each target environment before deploy. There is no `create` field; the CLI does not create projects.
 - You MUST include `workflow.create` in bootstrap actions for any deploy-target stage that has workflows.
-- You MUST follow the manifest schema in docs/manifest-schema.md (field names, types, required properties) as the source of truth.
+- You MUST follow the manifest schema (field names, types, required properties) as the source of truth — see the manifest schema doc linked in References.
 - You MUST validate that every `connectionName` in `content.storage`, `content.workflows`, and `deployment_configuration.storage` exists in the discovered connections.
 - You MUST add this exact exclude list to EVERY storage content item: `[".ipynb_checkpoints/", "__pycache__/", "*.pyc", ".libs.json"]`. When a storage item's `include` pattern would match the manifest itself (any `include` containing `*`, `*.yaml`, or the manifest's directory), you MUST also add `manifest.yaml` to that item's exclude list. Always append such extra excludes in addition to — never instead of — the four required patterns.
 - You MUST write `targetDirectory` values WITHOUT a trailing slash (`jobs`, not `jobs/`); the CLI appends the separator, so a trailing slash yields a malformed `//` path. Use `.` for the connection root.
@@ -45,7 +45,7 @@ This skill is READ-ONLY: it discovers project resources and presents generated c
 
 Skip if the user already specified exactly what to include. Use your agent's AWS access (e.g. the AWS CLI or an equivalent capability) to discover, scoping every call to the project's domain and project. Obtain the `domainId`/`projectId` from the user, or resolve them by listing (`aws datazone list-domains`, `aws datazone list-projects`).
 
-1. **Connections** — `aws datazone list-connections --domain-identifier <id> --project-identifier <id>`. Reveals S3, Glue, Athena, MWAA, MLflow, Redshift, EMR. See docs/connections.md.
+1. **Connections** — `aws datazone list-connections --domain-identifier <id> --project-identifier <id>`. Reveals S3, Glue, Athena, MWAA, MLflow, Redshift, EMR. See the connections doc linked in References.
 2. **Storage structure** — `aws s3 ls` (or `list-objects-v2` with delimiter `/`) on the shared S3 connection to find prefixes/file types: `src/` (`.py`), `workflows/` (`.yaml`), `data/`, `models/`, `notebooks/` (`.ipynb`).
 3. **MWAA Serverless workflows** — list workflows and post-filter by project tag `AmazonDataZoneProject`.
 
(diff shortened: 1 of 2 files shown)


skills/generate-bundle-manifest/SKILL.md
```diff
@@ -0,0 +1,211 @@
...
+- All projects referenced in the manifest MUST be pre-created by the user in each target environment before deploy. There is no `create` field; the CLI does not create projects.
+- You MUST include `workflow.create` in bootstrap actions for any deploy-target stage that has workflows.
+- You MUST follow the manifest schema (field names, types, required properties) as the source of truth — see the manifest schema doc linked in References.
+- You MUST validate that every `connectionName` in `content.storage`, `content.workflows`, and `deployment_configuration.storage` exists in the discovered connections.
+- You MUST add this exact exclude list to EVERY storage content item: `[".ipynb_checkpoints/", "__pycache__/", "*.pyc", ".libs.json"]`. When a storage item's `include` pattern would match the manifest itself (any `include` containing `*`, `*.yaml`, or the manifest's directory), you MUST also add `manifest.yaml` to that item's exclude list. Always append such extra excludes in addition to — never instead of — the four required patterns.
```
vasu2856, line 29: shouldn't we carry forward the libs.json to replicate the dependencies for jupyter notebooks?

skills/generate-bundle-manifest/SKILL.md
```diff
@@ -0,0 +1,211 @@
...
+- You MUST write `targetDirectory` values WITHOUT a trailing slash (`jobs`, not `jobs/`); the CLI appends the separator, so a trailing slash yields a malformed `//` path. Use `.` for the connection root.
+- Include a notebooks storage item when `.ipynb` files are discovered in the `notebooks/` folder (same as how Glue scripts are included when `.py` files are found). These are user-created notebooks used with `SageMakerNotebookOperator` or JupyterLab. If no `.ipynb` files are found, omit the notebooks storage item.
+- Catalog is OPT-IN. You MUST NOT set `content.catalog.enabled: true` unless the user explicitly asks. Catalog import needs `datazone:AddPolicyGrant` and ABORTS the whole deploy if the role lacks it. When you omit it, tell the user it can be enabled if they have DataZone catalog admin permissions.
+- You SHOULD use `append: true` for workflow storage items and omit `append` for source code.
+- You MUST set `applicationName` to match `^[A-Za-z0-9][A-Za-z0-9_-]*$` (max 50 chars). Derive it deterministically:
+  1. Strip a leading environment prefix if present: `dev-`, `test-`, `prod-`, `dev_`, `test_`, `prod_`.
```
vasu2856: Where will it get this 'environment prefix' from? shouldn't this be the name of the project instead?

skills/generate-bundle-manifest/SKILL.md
```diff
@@ -0,0 +1,211 @@
...
+
+The response lists the project's connections and their types (S3, Glue/SPARK, ATHENA, WORKFLOWS_MWAA, WORKFLOWS_SERVERLESS, MLFLOW, REDSHIFT, EMR). Record each connection NAME (e.g. `default.s3_shared`, `default.workflow_serverless`) — these are the values that go into the manifest's `connectionName` fields. For the shared S3 connection, read its S3 URI from `props.s3Properties.s3Uri` (the bucket is that URI with `s3://` stripped and everything after the first `/` removed). You need this URI for step 2.
+
+**2. Storage structure (uses the S3 URI from step 1).** List the shared S3 location to find top-level prefixes and the file types in each:
+
+```bash
+aws s3 ls <s3Uri-from-step-1>/ --recursive
+```
+
+Classify what you find: `.py` under `src/`, `src/glue-jobs/`, or `jobs/` → Glue/code; `.yaml` under `workflows/` → workflow DAGs; `.ipynb` under `notebooks/` → notebooks; `data/`, `models/` → data/model artifacts.
```
vasu2856: src? aren't these under shared?

skills/generate-bundle-manifest/SKILL.md
```diff
@@ -0,0 +1,211 @@
...
+
+**3. MWAA Serverless workflows** (only if the project has a `WORKFLOWS_SERVERLESS` connection):
```
vasu2856: doesn't every project have MWAA serverless connection by default?

skills/generate-bundle-manifest/SKILL.md
```diff
@@ -0,0 +1,211 @@
...
+| `.py` scripts in `src/` (or `src/glue-jobs/`, `jobs/`) | `content.storage` `name: code` |
+| `.ipynb` files in `notebooks/` | `content.storage` `name: notebooks` (include `notebooks/`) |
+| `.yaml` DAGs in `workflows/` | `content.storage` `name: workflows`, `append: true` |
+| Data in `data/` | `content.storage` `name: data` |
+| Model artifacts in `models/` | `content.storage` `name: models` |
+| Each discovered/generated workflow | one `content.workflows[]` entry |
+| Catalog (only if user opts in) | `content.catalog.enabled: true` |
+| QuickSight / Git (user-specified) | `content.quicksight[]` / `content.git[]` |
+
+`.ipynb` files co-located with a DAG in `workflows/` run via `SageMakerNotebookOperator` inside that DAG — keep them in the workflows storage item. Standalone `.ipynb` files (no DAG) are just replicated, no workflow needed. For SageMaker script-mode code (`sagemaker_submit_directory`), set `compression: tar.gz` on the deployment_configuration storage item.
```
vasu2856: how are ipynb and workflows related? workflows should be driven by presence of DAGs in dags folder or yaml files.

skills/generate-bundle-manifest/SKILL.md
```diff
@@ -0,0 +1,211 @@
...
+
+### Step 4: Deployment Configuration per Stage
+
+For each stage set: `domain` (dev: real ID/name; test/prod: `${STAGE_DOMAIN_NAME}`/`${STAGE_DOMAIN_ID}`), `region` (dev: real; test/prod: `${STAGE_DOMAIN_REGION:us-east-1}`), `project.name` (dev: real; test/prod: `${STAGE_PROJECT_NAME:default}`), `deployment_configuration.storage` (map each content item to a `targetDirectory`, no trailing slash), and `environment_variables`.
+
+`bootstrap.actions` — test/prod only, never dev:
+- Add ONE `workflow.create` (no `workflowName`) — it creates ALL workflows listed in `content.workflows[]`. Do not list names again or add one action per workflow.
+- Add `workflow.run` ONLY for a generated orchestration workflow (its Glue jobs are created by running it). Pre-existing replicated workflows get created but not run.
```
vasu2856, line 113: this seems specific to the example we have. running a workflow should be an opt in for the customer, as there is no guarantee that we will have a deployment workflow all the time. We should atleast as/confirm from the user that they want the workflows to be included.

skills/generate-bundle-manifest/SKILL.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+### Step 5: Generate Orchestration Workflow (when required)
+
+Before generating, check whether each Glue script is already referenced: read each discovered DAG and inspect every `GlueJobOperator.script_location`; a match (by file name or S3 path) means that script is covered.
+
+- If every Glue script is already referenced → do NOT generate a workflow; just include the scripts in a storage item.
+- Otherwise → generate exactly ONE orchestration workflow containing a `GlueJobOperator` task for each UNREFERENCED script (never one workflow per script), register it in `content.workflows[]`, and add a `workflow.run` action.
+
+Generate an orchestration workflow when: (a) unreferenced Glue/VETL scripts exist, (b) the user explicitly asks for a post-deployment workflow, or (c) multiple `.ipynb` files exist in S3 that the user has indicated must run in a defined order.
```
vasu2856: having .ipynb files doesn't mean we need to build a workflow.

skills/generate-bundle-manifest/SKILL.md
```diff
@@ -0,0 +1,211 @@
...
+
+Glue job replication: `GlueJobOperator` with `update_config: true` and `create_job_kwargs` creates-or-updates AND runs the job in one operation (`script_location` points to the deployed S3 path) — there is no separate create-job call, which is why generated Glue workflows MUST include `workflow.run`.
+
+When you generate a workflow, produce the actual DAG YAML in the response (do not just offer to). Default to MWAA Serverless YAML unless the user asks for provisioned (Python DAG). Use `{proj.connection.default.s3_shared.s3Uri}`, `{proj.iam_role_name}`, `{domain.region}` for substitution. For operator names, parameters, and DAG-authoring rules, consult:
+- The AWS operators/sensors available in SMUS workflows — https://github.com/aws/CICD-for-SageMakerUnifiedStudio/blob/main/docs/airflow-aws-operators.md
+- MWAA Serverless user guide — https://docs.aws.amazon.com/mwaa/latest/mwaa-serverless-userguide/what-is-mwaa-serverless.html
+
+Orchestration-workflow correctness is best-effort; the essential deliverable is a clean, valid manifest.
+
+### Step 6: Present Output (no writes)
```
vasu2856, line 159: We should also introduce dry run to the agent to help validate the manifest that's generated.

skills/generate-bundle-manifest/SKILL.md
Shnekit: Yes, missed it

Shnekit: Actually, I think these files should be excluded because they are just the dependencies of the local dev environment and we should probably not mess with test and production environments. They are also excluded in the examples in this repo and this is a general recommendation across the board (googled to confirm).

skills/generate-bundle-manifest/SKILL.md
Shnekit: I believe this is talking about the application name (the name of pipeline which goes at the top of the manifest). Not sure where the prefixes are coming from. But usually the agent ends up with something like this for a default name `AdminProject993760062230Pipeline`

skills/generate-bundle-manifest/SKILL.md
Shnekit: I think here it is just ensuring that if the glue jobs are put anywhere from `shared/src`, `shared/src/glue-jobs`, or `shared/jobs` (default location) we are able to find them. I can just leave `shared/jobs` directory, but I do not think it hurts leaving it

skills/generate-bundle-manifest/SKILL.md
Shnekit: Yeah, I think so. Fixed

skills/generate-bundle-manifest/SKILL.md
Shnekit: I think this was a left-over statement from my previous edits. Removed it.

skills/generate-bundle-manifest/SKILL.md
Shnekit: So, I think it is buried somewhere in the skill, and I actually tested this scenario on DataAgent, so it should be accurate:

By default, the skill will write the manifest with `workflow.create` for all existing workflows. And it will wait for user to say whether they want these workflows executed. If they say nothing, they will just be created.

There is a separate orchestration workflow that this skill can create and run if, for example, it finds glue jobs not connected to any existing workflow which can only be created through GlueJobOperator. 

So, by default, existing customer workflows are only created, but not run, and orchestration workflow (if exists) will be created and run.

skills/generate-bundle-manifest/SKILL.md
Shnekit: This is a bit confusing, yeah. I will remove point c.

skills/generate-bundle-manifest/SKILL.md
Shnekit: You mean to advise the customer to do a dry-run? It's already running as part of regular deploy, so I think it is okay?

skills/generate-bundle-manifest/SKILL.md
vasu2856: okay. Do we need that to be explicit somewhere?

Shnekit: Lines 112-113 say that `workflow.create` should be used with all existing customer workflows, `workflow.run` is used only with orchestration workflow. I think it is clear enough, but I can add a separate statement that existing workflows should not be run. But I think "ONLY" part takes care of that too.

skills/generate-bundle-manifest/SKILL.md
vasu2856: The applicationName should pick up the project name as default. Otherwise, it might look auto generated. It seems like stemming from our examples where we are prefixing the projects with dev-marketing, prod-marketing, etc

skills/generate-bundle-manifest/SKILL.md
vasu2856: We should let the agent look into shared to figure it out as there is no gaurantee 'src'

skills/generate-bundle-manifest/SKILL.md
vasu2856: We might have to test this out to confirm

skills/generate-bundle-manifest/SKILL.md
vasu2856: nope. So we are introducing the command suggestions to the agent to propose as next steps. Explicit dry run should be one of them, so that it naturally leads the customers through the process of 1. creating a manifest 2. create bundle 3. dry run 4. deploy

skills/generate-bundle-manifest/SKILL.md
Shnekit: Okay, no problem. Let me correct this.

skills/generate-bundle-manifest/SKILL.md
Shnekit: Fixed

skills/generate-bundle-manifest/SKILL.md
Shnekit: I did test it with DataAgent, it worked fine with 2 workflows and 1 orchestration workflow creating a GlueJob. I will also be testing it out more with Kiro and Claude to simulate exactly what customers might be seeing with their agent

skills/generate-bundle-manifest/SKILL.md
Shnekit: Okay, let's do that

vasu2856 (approved):
can you add a backlog to add evals to this? all of the AWS skills have evals associated with them. We should definitely add some here.

Shnekit:
Evals backlog: https://app.asana.com/1/8442528107068/project/1211853550096664/task/1217949781424750?focus=true

PR #353, opened 2026-09-15
fix(deploy): skip .ipynb_checkpoints in S3 DAG discovery

pupumaru:
## Problem
`_find_dag_files_in_s3` scans `*.yaml`/`*.yml` under the search prefixes without excluding `.ipynb_checkpoints/`. JupyterLab creates checkpoint copies of workflow YAMLs in the project's shared S3 prefix when edited in the portal; a checkpoint with the same top-level key / `dag_id` can be selected, registering a stale workflow definition.

## Fix
Skip S3 keys containing `.ipynb_checkpoints/` (and `__pycache__/`) in the object loop, consistent with the exclude patterns already documented for `content.storage`.

## Testing
- Added `tests/unit/commands/test_deploy_dag_discovery_skips_checkpoints.py` (real vs checkpoint copy; asserts the checkpoint is neither returned nor fetched).
- Existing `tests/unit/commands/test_deploy_*` pass (no regression).

Verified against the current `main` (matches v1.0.8).


Shnekit:
**Related gap: `_resolve_and_upload_workflows` has the same bug and isn't covered here**

Nice fix. While reviewing, I traced the other S3-listing / YAML-matching paths and found a sibling function in this same file with the identical issue that this PR doesn't touch: `_resolve_and_upload_workflows` (~line 694).

It lists `*.yaml`/`*.yml` under the prefix, runs `_is_workflow_yaml()`, then **resolves variables and re-uploads the file in place** — with no `.ipynb_checkpoints/` / `__pycache__/` guard:

```python
response = s3_client.list_objects_v2(Bucket=s3_bucket, Prefix=s3_prefix)
for obj in response["Contents"]:
    s3_key = obj["Key"]
    if not s3_key.endswith((".yaml", ".yml")):
        continue
    ...
    s3_client.upload_file(temp_file.name, s3_bucket, s3_key)  # rewrites in place
```

A checkpoint copy that has `dag_id` + `tasks` and a top-level key in `workflow_names` would get resolved and rewritten — arguably higher impact than the discovery path fixed here, since it mutates the object rather than just selecting it. Suggest extending this PR with the same guard:

```python
if ".ipynb_checkpoints/" in s3_key or "__pycache__/" in s3_key:
    continue
```

Lower priority, for defense-in-depth: `dry_run/checkers/workflow_checker.py::_gather_workflow_files` scans `zf.namelist()` for workflow YAMLs without the same filter. It's mostly protected because `bundle.py` strips those artifacts when building the zip, so only worth it if you want it robust to bundles built another way.

Longer term the two S3 workflow-discovery loops (`_find_dag_files_in_s3` and `_resolve_and_upload_workflows`) probably want to share one helper so a fix to one can't miss the other — which is how this gap arose.

Shnekit (changes requested):
Let's add that second fix

pupumaru:
Thanks for the careful review — you're right that `_resolve_and_upload_workflows` had the same gap, and it's the more impactful one since it rewrites the object in place. Pushed an update:

**The second fix (required):** extracted a shared `_is_editor_or_build_artifact()` helper and applied it in both `_find_dag_files_in_s3` and `_resolve_and_upload_workflows`, so a checkpoint copy is now skipped in both the discovery and the resolve/re-upload paths. Centralizing it addresses your "longer term" point too — a fix to one loop can't miss the other anymore.

**Extra fix while I was in there:** `_resolve_and_upload_workflows` was listing S3 with a single `list_objects_v2` call (capped at 1000 objects), while `_find_dag_files_in_s3` already paginates. That means a workflow beyond the first page would silently keep its unresolved `{proj...}` placeholders. Switched it to `get_paginator` for parity, with a regression test for a workflow on the second page.

**Not included (happy to follow up if you'd prefer):** the `workflow_checker._gather_workflow_files` zip/local path — as you noted it's mostly covered by `bundle.py` stripping those artifacts, so I left it out to keep this PR focused, but noted it in the helper's docstring so the third path is discoverable.

Tests added: helper unit test + discovery-skip + resolve-skip + pagination. `black` / `isort` / `flake8` clean on `src/`.

I also squashed the branch to two focused commits (skip fix, then pagination fix). Since this is a fork PR the new run is sitting in `action_required` — could you approve the workflow run when you get a chance? Thanks!

Shnekit (approved):
LGTM

PR #354, opened 2026-09-15
fix(deploy): sanitize invalid chars (incl. spaces) in generated workflow name

pupumaru:
## Problem
`_generate_workflow_name` replaces only `-` with `_`. A DataZone project name containing a space (e.g. "Silver Layer Project") produces a workflow name with a space, which MWAA Serverless `CreateWorkflow` rejects (400). The only current workaround is renaming the project.

## Fix
Normalize each name component to the `[A-Za-z0-9_]` character set MWAA Serverless allows, via `re.sub(r"[^A-Za-z0-9_]", "_", ...)`. Hyphen behavior is preserved. (`import re` added at module top.)

## Note for maintainers
If the officially supported charset is narrower/wider than `[A-Za-z0-9_]`, please point me at the spec and I'll match it exactly (or switch to an early, explicit validation error naming the offending project name).

## Testing
- Added `tests/unit/commands/test_deploy_workflow_name_sanitize.py` (hyphen preserved, space sanitized, only allowed chars remain).
- `black` clean; existing deploy unit tests pass.

Verified against the current `main` (matches v1.0.8).


Shnekit (approved):
Looks good! Might be worth adding a log line informing the customer of the workflow name change. Unlikely, but we might have customers define 2 workflows: "My workflow" and "My-workflow" which would result in the same workflow and collide. Non-blocking though.

PR #361, opened 2026-09-17
Speed up workflow deploy: single-pass DAG scan, fail-fast, tag updates

Shnekit:
## Summary

Addresses three customer-reported issues with workflow deployment.

**1. Single-pass DAG scan (`_find_dag_files_in_s3`)**
Previously each manifest workflow re-listed the S3 prefix and re-downloaded/re-parsed every YAML until a match was found — O(workflows × files). A workflow with no matching YAML never hit the early-exit, so it forced a full exhaustive scan of the whole prefix. Now each YAML is downloaded and parsed exactly once into a `{identifier -> s3_key}` lookup, and workflows resolve via a dict lookup — O(files).

**2. Fail-fast on missing workflow YAML**
A manifest workflow with no matching YAML now raises `WorkflowYamlNotFoundError` (naming the offending workflows) instead of emitting a warning and continuing. This is a clear manifest error (typo or stale entry) that the user can fix in seconds, and it removes the pathological exhaustive-scan path.

**3. Tag value updates (`_ensure_workflow_tags`)**
Reconciliation was add-only: changing a tag value in the manifest did nothing (logged "All required tags already present"). It now adds missing tags **and** updates changed values in a single `tag_resource` call. Tags present on the workflow but absent from the desired set are intentionally left untouched — no deletes — to avoid clobbering manually-added / out-of-band tags (backwards-compatible).

## Testing

- New `tests/unit/test_deploy_dag_scan_and_tags.py`: covers top-level-key match, `dag_id` match, single-pass scan (each YAML downloaded once), fail-fast on missing YAML, and the four tag cases (add / update value / no-op / never-delete-unmanaged).
- Full unit suite: 1158 passed (4 hypothesis-dependent property-test modules excluded — pre-existing env gap, unrelated).
- Merged latest `main` (incl. the workflow-name sanitize fix) before raising; no conflicts.

## Not included

A related IdC-domain fix (adding the GitHub Actions role as a DataZone project member so `describe --connect` can list connections) was applied directly as a runtime change to the dev-idc project; test-idc and prod-idc already had the membership. That is infra state, not code, so it is not part of this PR.


src/smus_cicd/bootstrap/handlers/workflow_create_handler.py
```diff
@@ -379,14 +379,30 @@ def _ensure_workflow_tags(
         typer.echo(f"   ⚠️  Could not read tags for {workflow_arn}: {e}")
         return
 
-    missing_tags = {k: v for k, v in desired_tags.items() if k not in current_tags}
+    # Reconcile the desired tags: add keys that are missing and update keys
```
vasu2856, line 382: won't always adding the tags update them, if they exist? Why do we have to keep track of what changed and figure out if they are add or update?

Shnekit: The current logic was to check the keys and add ones that do not yet exist. That's why we were only supporting tag additions before this. This was also because we wanted to make sure that reserved tags (like DataZoneDomainId, ProjectId) are not overwritten by the customer. 
But if you think that's too complicated, I can add a blind tag addition (and see if it can automatically overwrite the tag if already exists)

vasu2856: yeah. We should just simplify and always add what's there in the manifest. obviously, retain the reserved tags check.

vasu2856 (commented):
LGTM

vasu2856 (approved):
LGTM

